PR #45, opened 2015-02-10
World class with creation params merging

ftomassetti:


create_ancient.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
ftomassetti: Should we have a Python script for that?

README.md
```diff
@@ -265,6 +265,8 @@ I would like to thank you great people who helped me while working on Lands:
 
 * [stefan-feltmann](https://github.com/stefan-feltmann) made Lands depends on pillow instead that on PIL (which is deprecated). This could also help when moving to Python 3. Thanks a million!
 
+* [Russell Brinkmann](https://github.com/rbb) helped saving the generation parameters in the generated world (so that we can use it to generate the same world again, for example), improved the command line options and added tracing information (useful for understanding the performance of the various generation steps)
```
ftomassetti: @rbb does it sound ok for you?

create_ancient.sh
rbb: It is just a quick throw together script to both create a world file and draw the ancient map.I think that longer term it goes away. Some how, the command line use of generator.py should allow complete creation of the world file and drawing the ancient map.

README.md
rbb: Works.

create_ancient.sh
ftomassetti: ok, then I will drop it for now. I recently added the scripts "lands", which can be used instead of python lands/generator.py. You need to run python setup.py install or python setup.py develop to use it

ftomassetti:
This is going to be version 0.5.3, some good stuff from @rbb 
Thanks a lot!

PR #48, opened 2015-02-13
Announcing the merge and adding reference to the google group

ftomassetti:

PR #52, opened 2015-02-18
Adapting the erosion and rivers calculation system from WorldSynth

ftomassetti:
Given it is much better of the one we had in Lands, let's just drop it as a replacement.

Note that we are producing but yet not using the river map and lake map, so some adjustment will be necessary.

While we are using Numpy during the calculations we are then producing a plain Python matrix. We should move to numpy but we should do that for all the layers at once (and decide how to serialize numpy arrays).

This branch is rebased on top of the one with all the renaming to WorldEngine, so this pull request should be discussed after merging #48


psi29a:
Holy crap! This is really too big to review properly.

What code is still non-MIT?

ftomassetti:
It is definitely too big, but consider that it is built on top of #48 (which should be just about renaming here and there: long but not technical). Once we have merged that this is going to be much shorter. I can also separate a couple of things in a different branch, which can be accepted separately from this one. In the future I will try to make simpler pull requests, my bad.

Maybe there is still some reference to the old license, that is because this derive from the first branch that I started working on before we merged the project. I will clean up this branch, but first we need #48

psi29a:
I kicked the travis-ci builds again, this time they ran without a problem. ;)

ftomassetti:
I think Travis was confused because we changed the name of the project (then I fixed the Travis configuration).

Linthub seems to like to complain... let's do this:
1) Approve #48
2) Do a branch exclusively for restyling
3) Reapply this branch on top of the previous 2, so we should have a way smaller diff and hopefully less complaints from linthub (not sure we can remove all of them).

How does it sound?

psi29a:
Sure, I'm not worried (nor anal-retentive) about linthub.

We're in transition.

psi29a:
code is no longer stolen, it is gifted ;)

PR #54, opened 2015-02-24
Cleanup

ftomassetti:
I just run the tool [pep8](https://pypi.python.org/pypi/pep8) against all the files.

The most annoying part was to limit lines at 80 columns


psi29a:
Much of the value of PEP-8 is to stop people arguing about inconsequential formatting rules, and get on with writing good, consistently formatted code.

I suggest that we adopt PEP-8 for many things , but not for line length. I think 100 or even 120 columns is best. 

We do that work because most people work using 16:9 widescreen.

psi29a:
Try running pep8 with --max-line-length=120  :)

ftomassetti:
```Try running pep8 with --max-line-length=120```

I wish I knew that a few hours ago :)

Do you think I should drop and redo the changes or it is fine anyway?
Definitely from now on I will use this flag!

psi29a:
It is fine as it is. ;)

We'll just make an agreement that every time we visit a file, we pep8 it with 120 line length.

ftomassetti:
Created coding guidelines suggesting to use pep8 with 120 line lenght: https://github.com/Mindwerks/worldengine/wiki/Coding-guidelines

Thanks!

PR #58, opened 2015-02-25
Trivial fix with Greyscale

ftomassetti:
We have now generic functions (like `draw_grayscale_heightmap`) which can be used both for the GUI and to write on image files. Then we have functions like `draw_grayscale_heightmap_on_file` which wraps them, executing the drawing on file.

We were using the wrong one in this case :(


psi29a:
Setting appveyor to skip branches without appveyor.yml

ftomassetti:
that makes a lot of sense, thanks :)

ftomassetti:
I do not recall what happened but this seems fixed in master, so I will close this one

PR #59, opened 2015-03-08
Lets create an output directory if it does not exist

cesarramirez:
For convenience purposes I suggest that we can create the output directory if it does not exist.


ftomassetti:
Thanks, it seems correct. You have also fixed the problem about generating greyscale maps that we introduced with a recent refactoring. Feel free to send more of these patches :D 

Thanks for your contribution, it is appreciated!

psi29a:
Great work, thanks! :)

PR #53, opened 2015-02-20
Appveyor

ftomassetti:
Fixed PyPlatec (and released), now WorldEngine can be built on AppVeyor

PR #67, opened 2015-05-04
fix #66: ancient_map load pickle or protobuf worlds

### LINKED ISSUE #66: Operation ancient_map load only worlds in Pickle format
```
    elif operation == 'ancient_map':
        # First, some error checking
        if options.sea_color == "blue":
            sea_color = (142, 162, 179, 255)
        elif options.sea_color == "brown":
            sea_color = (212, 198, 169, 255)
        else:
            usage("Unknown sea color: " + args[0] + "  Select from [" + SEA_COLORS + "]")
        if not options.world_file:
            usage("For generating an ancient map is necessary to specify the world to be used (-w option)")
        world = World.from_pickle_file(options.world_file)
```

ftomassetti:
In ancient_map operation we are not trying to load the world first as pickle file then as protobuf file, if an exception is thrown. This is probably not the best way to do it and maybe we could create a method name "load_world_file".


ftomassetti:
we should wait for the Appveyor pull request (#53) before merging it, so we are sure we are not breaking anything on windows.

ftomassetti:
About the exception: apparently we have no easy way to distinguish a pickle file from a protobuf file, so we just try first to load a file as a pickle file, if this fails we then try to load it as a protobuf file. When the file is a protobuf file an exception is thrown but we do not want to show that necessarily. Maybe we can store the exception and print it when we cannot load the file neither as a pickle or a protobuf file.

psi29a:
Is there a file header? Normally the first 4 bytes can be read to see what format it is. Not sure if this helps or not. :)

psi29a:
https://stackoverflow.com/questions/13939913/how-to-test-if-a-file-has-been-created-by-pickle
^-- pickle doesn't have a reliable header, but the last byte should be the pickle stop op-code we can test for. :)

psi29a:
Another trick, with photobuf is to always make the very first thing written to be a "version" 4-byte of your choice. This way, you can detect your own file format...  give it a try. That way you can test that the file you are inputting is something created by worldengine.

ftomassetti:
ok, we could try to use these 2 tricks (stop on code and protobuf headers). They seem to be not 100% exclusive but still in practice it should be enough.

ftomassetti:
We also:
* update the version number (we now use it in the protobuf format)
* we changed the protobuf format so we update the worlds used in tests

ftomassetti:
Rebasing after having merged appveyor

ftomassetti:
ok, we now should be able to recognize the file type (protobuf vs pickle). I do not understandthe error from linthub. @psi29a seems good to you?

psi29a:
I don't understand the error either, I've never used linthub before. ;)
The rest looks good!

ftomassetti:
yes, I am tempted to just remove linthub... for now I will merge the PR :)

PR #69, opened 2015-05-17
Create setup.py

SourceRyan:
Edited setup.py for Windows users, forcing them to use numpy 1.9.2 due to the lack of wheel files for earlier versions.

### DIFF
--- a/windows_binaries/setup.py
+++ b/windows_binaries/setup.py
@@ -1,0 +1,35 @@
+from setuptools import setup, Extension, Command
+# from pip.req import parse_requirements
+# import pip.download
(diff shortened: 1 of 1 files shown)


ftomassetti:
well, I guess we can just move to numpy 1.9.2 for every user, so changing both the setup.py and the requirements_base.txt file. @psi29a is the NumPy expert: is it ok for you?

psi29a:
Go ahead! :)

ftomassetti:
Ok, then I will merge #71 and close this one. Another time: Thank you Ryan!

Issue #75, opened 2015-06-13
Add an option to generate B/W temperature map


ftomassetti:
Add an option to generate B/W temperature map

ftomassetti:
Working on it in branch bw_temperature

ftomassetti:
Merged, closing

Issue #74, opened 2015-06-13
Generate B/W precipitation map


ftomassetti:
Add an option to generate B/W precipitation map

ftomassetti:
Working on it in branch bw_temperature

ftomassetti:
Merged, closing

PR #78, opened 2015-06-17
Black and white temperature and precipitations map

ftomassetti:
We add an option to generate black and white maps and we adapt temperature and precipitations drawing routines to consider that option.

See #74 and #75


psi29a:
Looks good and travis is happy :)

PR #86, opened 2015-06-24
Add info operation

ftomassetti:
The info operation print on screen basic information about a saved world file.

In the future we could add more information (like the average elevation of the number of cells for each biome type)


psi29a:
Looks good to me. :)

ftomassetti:
Great, merging!

PR #96, opened 2015-06-30
use argparse instead of optparse, fix running worldengine directly wi…

psi29a:
…thout first installing


worldengine/cli/main.py
```diff
@@ -400,79 +403,78 @@ def main():
...
-        operation_ancient_map(world, options.generated_file,
-                              options.resize_factor, sea_color,
-                              options.draw_biome, options.draw_rivers,
-                              options.draw_mountains, options.draw_outer_border)
+        if not args.generated_file:
+            args.generated_file = "ancient_map_%s.png" % world.name
+        operation_ancient_map(world, args.generated_file,
+                              args.resize_factor, args.sea_color,
+                              args.draw_biome, args.draw_rivers,
+                              args.draw_mountains, args.draw_outer_border)
```
ftomassetti: Thanks for refactoring all of this! Maybe later we could split this function in argument processing, validation and execution

worldengine/generation.py
```diff
@@ -1,14 +1,14 @@
...
 from worldengine.simulations.basic import find_threshold_f
-from worldengine.simulations.WatermapSimulation import WatermapSimulation
-from worldengine.simulations.IrrigationSimulation import IrrigationSimulation
-from worldengine.simulations.HumiditySimulation import HumiditySimulation
-from worldengine.simulations.TemperatureSimulation import TemperatureSimulation
-from worldengine.simulations.PermeabilitySimulation import PermeabilitySimulation
-from worldengine.simulations.ErosionSimulation import ErosionSimulation
-from worldengine.simulations.PrecipitationSimulation import PrecipitationSimulation
-from worldengine.simulations.BiomeSimulation import BiomeSimulation
+from worldengine.simulations.hydrology import WatermapSimulation
```
ftomassetti: +1 for the nicer names: I am learning python conventions

ftomassetti:
Looks good to me, Travis and AppVeyor!

tests/astar_test.py
```diff
@@ -0,0 +1,23 @@
+import unittest
+import numpy
+from worldengine import astar
+
+
+class TestCommon(unittest.TestCase):
+
+    def test_traversal(self):
```
ftomassetti: nice test. Related thing: it could be useful to have some documentation to understand how astar should be used. For examples, does 0.0 represents a point that cannot be traversed or it is something that can be traversed for free?

tests/cli_test.py
```diff
@@ -0,0 +1,56 @@
...
+from worldengine.cli.main import main
+
+
+class TestCLI(TestBase):
+
+    def setUp(self):
+        super(TestCLI, self).setUp()
+        self.world = "%s/seed_28070.world" % self.tests_data_dir
+
+    def test__main__(self):
```
ftomassetti: nitpick/pointless: I am not sure if we should include the underscores in the name. I am afraid someone could think this method is private because it terminates with underscores.

tests/cli_test.py
```diff
@@ -0,0 +1,56 @@
...
+        assert __main__
+
+    def test_options(self):
```
ftomassetti: +1

maybe we could split this one in a few separate tests, to verify each single operation/option works (of course I am not expecting you to write tests for all these cases as part of this PR)

tests/cli_test.py
```diff
@@ -0,0 +1,56 @@
...
+        sys.argv = ["python", "info"]
+        self.assertRaises(SystemExit, main)
+        sys.argv = ["python", "info", self.world]
+        try:
+            main()
+        except Exception as e:
+            raise e
+        # TODO: fill in the rest of the options and their possibilities
+
+    def test_smoke_full(self):
```
ftomassetti: I am very in favor of such tests. If we ever have a few we could move them in a separate collections of smoke tests

worldengine/cli/main.py
```diff
@@ -120,18 +120,20 @@ def check_step(step_name):
...
+    draw_ancientmap_on_file(world, map_filename, resize_factor, sea_color,
+                            draw_biome, draw_rivers, draw_mountains,
+                            draw_outer_land_border)
     print("+ ancient map generated in '%s'" % map_filename)
 
 
 def __get_last_byte__(filename):
-    with open(filename, 'rb') as ifile:
-        data = tmp_data = ifile.read(1024 * 1024)
+    with open(filename, 'rb') as input_file:
```
ftomassetti: thanks for cleaning this and the following fuction!

worldengine/cli/main.py
```diff
@@ -223,175 +227,180 @@ def print_world_info(world):
...
+    g_ancient_map.add_argument('--not-draw-rivers', dest='draw_rivers',
+                               action="store_false",
+                               help="Not draw rivers",
+                               default=True)
+    g_ancient_map.add_argument('--draw-outer-border', dest='draw_outer_border',
+                               action="store_true",
+                               help="Draw outer land border",
+                               default=False)
 
     # TODO: allow for RGB specification as [r g b], ie [0.5 0.5 0.5] for gray
```
ftomassetti: should this comment be moved?

ftomassetti:
looks good to me, just minor comments added (nothing blocking)

tests/astar_test.py
psi29a: You're right, there needs to be more documentation in astar.py, if you wish, create an issue otherwise I'll get to in time.

Astar works on cost, the higher the cost the less likely it is to travel that path.  There are no hard limits, it works on minus infinity and positive infinity. It will take a starting position and and end position, then find the path between the two with the lowest cost.  This is perfect for heightmaps for example, because you can use it to find path through mountain/hills between villages. ;)

tests/cli_test.py
psi29a: What do you mean?  There is no hard or fast rule on this, but typically in Python when you begin a function or variable/member with _ it is meant to declare it as only usable from within a particular scope. 

Now ipython will 'hide' those, but you can still access them if you really want to and most IDEs will complain (warnings in PyCharm). However in testing, it doesn't matter as you want to test ALL functions anyway.

Not sure if that answers your question or not.

ftomassetti: yes, the two underscore have this special meaning (just by convention) when present at the start of a name, but I thought it could be confusing to have them at the end (someone could wonder if they have the same meaning of the underscores at the start). But it was just over-worrying, let's drop my comment

PR #102, opened 2015-07-07
more coverage

psi29a:


psi29a:
awesome!  ;)  I get us updated to PyQt5 and suddenly coveralls is happy... whatever, I'll take it! :D

tests/cli_test.py
```diff
@@ -26,6 +27,20 @@ def test_options(self):
         self.assertRaises(SystemExit, main)
         sys.argv = ["python", "info", "does_not_exist"]
         self.assertRaises(SystemExit, main)
+        sys.argv = ["python", "info", "--gs", "greyscale.png"]
```
ftomassetti: good, I like more coverage. Maybe we could add messages in the asserts to explain why we expect those checks to fails (e.g., number-of-plates 101 should fail because the limit is to 100)

worldengine/gui/main.py
```diff
@@ -2,12 +2,15 @@
...
+from PyQt5 import QtCore, QtGui
+from PyQt5.QtGui import QImage, QPixmap
+from PyQt5.QtWidgets import QApplication, QDialog, QMainWindow, QAction, \
+    QFileDialog, QLabel, QWidget, QGridLayout, QPushButton, QLineEdit, QSpinBox
 import platec
 import random
 import sys
 import threading
-from world import World, Step
+from worldengine.world import World, Step
```
ftomassetti: it was broken before then, I guess

worldengine/gui/main.py
```diff
@@ -25,47 +28,47 @@
 from worldengine.views.WatermapView import WatermapView
 
 
-class GenerateDialog(QtGui.QDialog):
+class GenerateDialog(QDialog):
```
ftomassetti: thanks for removing all the QtGui: quite repetitive :)

ftomassetti:
LGTM! :shipit:

PR #105, opened 2015-07-10
removing the GUI

psi29a:


ftomassetti:
I thought we implemented more views...

Anyway, it seems good to me :ship: it!

psi29a:
show me your views!

psi29a:
86%  now we can badge it ;)

PR #108, opened 2015-08-18
Export

psi29a:
Adds support to export the heightmap from the world file into any format supported by libgdal.


CHANGELOG.md
```diff
@@ -1,24 +1,32 @@
...
+Version 0.19-UNRELEASED
 
-* Evan Sampson added Holdridge life zones model
+* Added export support using libgdal
 
-Version 0.4.1
+Version 0.18
 
-* Generation of grayscale heightmap
+* Merged Lands and WorldSynth into one WorldEngine
```
ftomassetti: thanks for filling this!

CHANGELOG.md
```diff
@@ -1,24 +1,32 @@
...
+* River and its erosion are now taken into account
+* Total restructuring of code including many bug fixes.
+* Windows/OSX/Linux binary builds
 
-Version 0.4.2
+Version 0.5.4
```
ftomassetti: We should probably remove everything related to lands alone

worldengine/cli/main.py
```diff
@@ -9,13 +9,14 @@
     draw_precipitation_on_file, draw_grayscale_heightmap_on_file, draw_simple_elevation_on_file, \
     draw_temperature_levels_on_file, draw_riversmap_on_file
 from worldengine.plates import world_gen, generate_plates_simulation
+from worldengine.imex import export
```
ftomassetti: imex stands for import/export? It is not super clear to me. Maybe we could call it importexport?

worldengine/cli/main.py
```diff
@@ -335,9 +336,24 @@ def main():
...
                                help="Draw outer land border",
                                default=False)
-
     # TODO: allow for RGB specification as [r g b], ie [0.5 0.5 0.5] for gray
 
+    # -----------------------------------------------------
+    export_options = parser.add_argument_group(
+        "Export Options", "You can specify the formats you wish the generated output to be in. ")
+    export_options.add_argument("--export-type", dest="export_type",
+                                help="Export to a specific format such as: BMP or PNG",
```
ftomassetti: does it mean we support these formats and only these formats?

worldengine/cli/main.py
```diff
@@ -335,9 +336,24 @@ def main():
...
+                                default="bmp")
+    export_options.add_argument("--export-bpp", dest="export_bpp", type=int,
+                                help="Bits per pixel: 8, 16 and 32",
+                                default=8)
+    export_options.add_argument("--export-signed", dest="export_signed", action="store_true",
```
ftomassetti: we should probably explain what happens when we export with this option on or off

ftomassetti:
We should probably describe the new option in the README (or move that part to a separate MANUAL file). I think it is worthy to tell at the top of the README that generated worlds can be exported.

AppVeyor is failing :(

ftomassetti:
"don't know how to compile Fortran code on platform 'nt'"

CHANGELOG.md
psi29a: I did, just re-ordered so that it reads top to bottom with the latest being on top. It was original bottom to top.

worldengine/cli/main.py
psi29a: imex is used other higher profile projects, I've not seen any that have used importexport and I'm not particularly fond of that either. ;)

worldengine/cli/main.py
psi29a: "such as" just gives you examples... I can't give you a full list because it would take up pages ;)

worldengine/cli/main.py
psi29a: It is in the help text, if you don't know what it means... then best to stay away from it. ;)

Besides, if you export to a format that doesn't supported signed ints, then it will yell at you anyway.

CLI isn't for dummies, that is what the GUI is for. The target audience of the CLI will likely know what signed means.

psi29a:
There will be a new manual section, using sphinx to do not only API documentation but also holding manual for a more verbose explination. This also includes screenshots and most of the stuff int he huge README.md file

psi29a:
appveyor is going to take time, first we need to find out how to install libgdal on windows :)

psi29a:
Maybe we can try this:
https://trac.osgeo.org/osgeo4w/wiki

CHANGELOG.md
ftomassetti: I meant: we should remove these references to versions < 0.18.0 because they are not related to WorldEngine

worldengine/cli/main.py
ftomassetti: Ok, cool :)

psi29a:
Yay, containers!  travis is now super responsive, we now have 4GiB of ram instead of 3, 2 dedicated cores instead of 1.5...   now to fix pypy

worldengine/cli/main.py
```diff
@@ -335,9 +336,24 @@ def main():
...
+                                help="Used signed bits or not.",
```
LavenderMoon: "Used signed bits or not." sounds incredibly clunky. "Use signed integers?" would make more sense to users, assuming I'm not missing something.

psi29a: Not really, because I plan on supporting floats as well. ;)  Using Int makes sense for visual formats like png and bmp, but we'll need to support float to keep precision. GTiff for example supports float as do most other DEM formats.

How does this sound:  "Use signed output (values can be negative), as opposed to unsigned (values are only positive)."

LavenderMoon: Oh, right, I didn't consider that. That new description is absolutely perfect.

ftomassetti: By the way is nice to have someone else taking a look at WorldEngine, thanks @LavenderMoon 

And if you want to contribute we are always looking for help :D

LavenderMoon: I'd love to help with WorldEngine and Civs, but I'm not much good with Python or Clojure. I'll be helping in any way I can, though. Thanks for the kind words, by the way. :D

psi29a: And with C? :)  There are various c libs (with python bindings) that we use that can be improved, for example the plate tectonic side of things.

Anyway, glad you eye-balling things here. Thank you!

LavenderMoon: Hmm, that's a fair point. While I'm absolutely abysmal with C++, C is a different story; well written C is usually fairly simple, though slow, for me to grasp. I'll look that those sometime soon.

PR #111, opened 2015-09-19
ECS

esampson:
Adjusted curves in humidity and temperature simulations so that groups correspond closer to Earth norms.

Added some weighting variables in humidity simulation to make it easier to adjust output. Increased the weighting of latitude for determining temperature in temperature simulation to produce a result that seems closer to Earth's (less chaotic).

Added an error catch to astar.py to prevent a CTD error that seemed to occur on maps of 1024x512 and larger.

Added a variable called latshift to generation.py. This will ultimately allow people to rerun a specific seed but shift the continental position up and down. This way if you identify a spot you think would be a good central location but you think that the campaign/story would benefit from having it further north or south the simulation can be rerun and the results shifted.

Reversed the order of generating temperature and humidity in generation.py as the temperature has an effect on the humidity.

Added biomes.scm, a script for GIMP that should create layers and layer groups for the biome png files.


psi29a:
Any way you can add some coverage by writing tests?

Unit tests for anything you can reproduce in a function level.
Integration tests for seeing that nothing breaks when running with command line arguments.

psi29a:
Thank you for the A* catch! :D

ftomassetti:
Thank you so much for sending this! And the script for Gimp is a great idea: we should consider exporting all our data to Gimp.

esampson:
I don't think we want to move the Data to any form of specific graphics software. The advantage of PNG is that people can use pretty much any software. I like GIMP because I've gotten pretty good with it an I like supporting Open Source software but I understand that some people find it more cumbersome to use (I still can't get my wife to use is and it drives me nuts over the crummy software she does use when she wants to print images).

I notice that my pull request is failing the coverage check. This is because if I try and use the must up to date branch of WorldEngine I get an error. I could fold the changes I've made into the latest release but I wouldn't be able to test them on my own (it looks like they should be fine because they occur in areas completely unrelated to the changes between 0.18 and current) so let me know if you want me to incorporate them and resubmit or what.

I can look into writing some unit tests but to be honest that isn't something I've got any experience in so it may take me a little bit to get up to speed with that.

psi29a:
I'm not sure what branch you are talking about, but under normal circumstances, your (or anyone's) pull request should always be against master.  

I think what is really happening is that your A* fix is preventing erosion from running. To test this, can you revert your A* fix and push? I think coveralls will show us a different picture then. 

If my hunch is correct, then you needn't have to write unit tests... but you should probably read through ours and see how we do testing. It helps us catch problems. :)

esampson:
Ok. I updated to the latest base and backed out the astar change and it is still failing the coverage check.

psi29a:
Super, -0.4% is better than -4.0% in my books! :)


Can you please create a new issue about the A* problem and how I can duplicate the issue?

psi29a:
@ftomassetti I'm OK with this as-is, and you?

ftomassetti:
Me too: I took a look at the changes yesterday and the code seems fine. I do not understand the GIMP script so... I will trust @esampson :)

Feel free to merge!

ftomassetti:
@esampson you are already listed among the contributors but let us know if you want us to update the description of your contributions

Issue #116, opened 2015-09-22
"--ocean-level" parameter didn't seem to do anything


ftomassetti:
I also remember that the "--ocean-level" parameter didn't seem to do anything. Any value down to 0.0001 (the generation cancelled with a value of 0) resulted in approximately the same amount of blue on the ocean map (although I might misunderstand what the parameter is actually supposed to do).

(reported by Stephan)

tcld:
I already created my very own issue for this, but your help is appreciated. :)
https://github.com/Mindwerks/worldengine/issues/114

I included some ocean maps there to provide a quick overview of the problem.

PS: I am the person you just exchanged e-mails with.^^

tcld:
This could technically be closed since it doesn't really contain any more information than #114 .

Issue #114, opened 2015-09-21
parameter --ocean_level might not have an influence


tcld:
Thanks to Frederico I just noticed I could open up issues. So I will:

Using the binary Linux build I noticed that the paramter "--ocean-level" does not seem to have any influence. I might be misunderstanding it, though. At the end of this post are some extreme examples: Two worlds with a value of 1x10e-5, two with 1.0 and two with 1x10e6. If any of the other maps would be useful in conjunction with this, let me know.

![10e-5_ocean1](https://cloud.githubusercontent.com/assets/12257447/9999077/af56f690-6094-11e5-8ea4-b88cca145dc9.png)
![10e-5_ocean2](https://cloud.githubusercontent.com/assets/12257447/9999079/af65bb12-6094-11e5-93f0-962e40f3faed.png)
![10e 1_ocean1](https://cloud.githubusercontent.com/assets/12257447/9999078/af5d9950-6094-11e5-87e5-e41a60e52e66.png)
![10e 1_ocean2](https://cloud.githubusercontent.com/assets/12257447/9999080/af66c52a-6094-11e5-8c9d-521c6f6764e8.png)
![10e 6_ocean1](https://cloud.githubusercontent.com/assets/12257447/9999082/af6b86aa-6094-11e5-805c-81eb7b07aaf2.png)
![10e 6_ocean2](https://cloud.githubusercontent.com/assets/12257447/9999081/af67c682-6094-11e5-8a4b-7f7fc686641d.png)

tcld:
I took a shot at this: https://github.com/Mindwerks/worldengine/pull/125
Wasn't able to test it myself, though, since I am still restricted to running the latest worldengine release instead of more recent versions.

psi29a:
How do you mean? Latest release or from source (master)?

tcld:
I worked on master, but am only able to run the release version from here:
https://github.com/Mindwerks/worldengine/releases

I never figured out how to run the non-binary version.

psi29a:
Did you try following the directions in the readme?

tcld:
Thanks for the advice. I tried many things, but I guess my Python installation is somehow messed up. I can always fall back to using a fresh installation inside a virtual machine, though.

psi29a:
No need, that is beauty of virtualenv, everything happens inside your sandbox.

PR #125, opened 2015-10-12
Parameter ocean_level was not passed to platec.

tcld:
I took a look into the problem of the inability to create oceans with differing depths.
According to the following comment from https://github.com/Mindwerks/plate-tectonics/blob/master/src/lithosphere.cpp
    // Find the actual value in height map that produces the continent-sea ratio defined be "sea_level".

I assume that said sea_level is indeed what is responsible for the depth of the ocean and hence isomorphic to the variable ocean_level used in https://github.com/Mindwerks/worldengine/blob/master/worldengine/plates.py

That variable, however, was not passed to generate_plates_simulation() and thus not to platec (line 42 of the commit).

I am still unable to run the worldengines source-code, therefore I cannot guarantee that this fixes the problem described here:
https://github.com/Mindwerks/worldengine/issues/114
https://github.com/Mindwerks/worldengine/issues/116

At the very least this looks like a sensible change and one that is simple enough to not cause problems.

This is my very first pull request. Feedback on its style is appreciated.

### DIFF
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -297,7 +297,7 @@ def main():
                             dest='grayscale_heightmap', action="store_true",
                             help='produce a grayscale heightmap')
     g_generate.add_argument('--ocean_level', dest='ocean_level', type=float,
-                            help='elevation cut off for sea level " +'
+                            help='amount of surface covered by ocean " +'
                                  '[default = %(default)s]',
                             metavar="N", default=1.0)
 
--- a/worldengine/generation.py
+++ b/worldengine/generation.py
@@ -98,16 +98,18 @@ def fill_ocean(elevation, sea_level):
 
     ocean = [[False for x in range(width)] for y in range(height)]  # TODO: use numpy
     to_expand = []
-    for x in range(width):
+    for x in range(width):#handle top and bottom border of the map
         if elevation[0][x] <= sea_level:
             to_expand.append((x, 0))
         if elevation[height - 1][x] <= sea_level:
             to_expand.append((x, height - 1))
-    for y in range(height):
+    for y in range(height):#handle left- and rightmost border of the map
         if elevation[y][0] <= sea_level:
             to_expand.append((0, y))
         if elevation[y][width - 1] <= sea_level:
             to_expand.append((width - 1, y))
+
+    #generate the ocean, starting on borders and then moving inwards
     for t in to_expand:
         tx, ty = t
         if not ocean[ty][tx]:
@@ -119,24 +121,30 @@ def fill_ocean(elevation, sea_level):
     return ocean
 
 
-def initialize_ocean_and_thresholds(world, ocean_level=1.0):
+def initialize_ocean_and_thresholds(world, ocean_level=0.65):
     """
     Calculate the ocean, the sea depth and the elevation thresholds
     :param world: a world having elevation but not thresholds
-    :param ocean_level: the elevation representing the ocean level
+    :param ocean_level: the amount of surface area covered by ocean
     :return: nothing, the world will be changed
     """
     e = world.elevation['data']
-    ocean = fill_ocean(e, ocean_level)
+
+    #Calculate the height of the ocean relative to the geometry
+    ocean_maxlevel = find_threshold_f(e, 1.0 - ocean_level,
+                                      ocean=None, max=1.0, mindist=0.00001)
+    #end of calculation    
+    
+    ocean = fill_ocean(e, ocean_maxlevel)
     hl = find_threshold_f(e, 0.10)
     ml = find_threshold_f(e, 0.03)
-    e_th = [('sea', ocean_level),
+    e_th = [('sea', ocean_maxlevel),
             ('plain', hl),
             ('hill', ml),
             ('mountain', None)]
     world.set_ocean(ocean)
     world.set_elevation(e, e_th)
-    world.sea_depth = sea_depth(world, ocean_level)
+    world.sea_depth = sea_depth(world, ocean_maxlevel)
 
 
 # ----
--- a/worldengine/plates.py
+++ b/worldengine/plates.py
@@ -11,7 +11,7 @@ from worldengine.common import array_to_matrix
 from world import World
 
 
-def generate_plates_simulation(seed, width, height, sea_level=0.65,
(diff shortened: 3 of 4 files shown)


tcld:
One thing left to mention:
I renamed the parameter "sea_level" in the Python code to make it consistent. That might be considered bad style, though.

psi29a:
Looks like the smoke-test went up in flames.

```
======================================================================
ERROR: test_smoke_full (tests.cli_test.TestCLI)
----------------------------------------------------------------------
  File "/home/travis/build/Mindwerks/worldengine/tests/cli_test.py", line 61, in test_smoke_full
    raise e
RuntimeError: maximum recursion depth exceeded in cmp
```

Why did you just rename the sea_level variable to ocean_level?
Why did you set the default value to 1.0?

tcld:
I renamed the variable for the sake of consistency. The same goes for the default value - ocean_level is handed down through half a dozen functions, all of which default to 1.0, and as long as it is set anywhere along that chain the final default value will not have an influence anyway.

@psi29a I noticed that your commit didn't pass that final check at first - is there any trick to apply?

tcld:
I now understand. sea_level refers to the ratio of the surface-areas of land and sea. Terribly named variable.

What exactly does ocean_level refer to and, if it is something different, how to translate it to sea_level?

psi29a:
What do you mean, my commit? This is your commit. :)

I suggest running the tests on your machine locally until you find out what the problem is and fix it. It is what we do and why we have tests to begin with. Unless of course, the test is faulty somehow. ;)

psi29a:
@ftomassetti this is your field, what does ocean_level mean and how does it relate to sea_level?

tcld:
I think I know what the problem is:

So:
ocean_level as used in the worldengine-code refers to some height relative to the geometry of the world. (Which is a bit problematic since the meaning of a value of "1.0" is seems mysterious.)
sea_level refers to the amount of surface covered by oceans, and that is how platec uses it.

The two are very similar, and there actually is a method in simulations/basic.py that could probably be used for mapping one to the other.

Which course of action would be the better one now: Completely eliminating ocean_levels meaning in favour of the sea_level one (new name: e.g. ocean_coverage), or mapping ocean_level to sea_level when communicating with platec? The latter way seems really difficult to implement, the first one should come with renaming a variable, which might be bad.

psi29a:
Normally, 0.0 means the lowest possible float point (when dealing with positive numbers) and 1.0 is the highest possible value. So basically the heightmap is made of floating points between 0.0 and 1.0
so a sea-level of 1.0 would mean that everything is sea! :)

tcld:
I would have expected something like "1.0 refers to an ocean level that is 1 meter above the lowest point of geometry". That would allow for an ocean that covers the whole planet (and even goes above the highest mountains).

I am now looking into replacing ocean_level with sea_level. (And into virtualenv, thank you. :) )

psi29a:
Sadly we can't deal in infinity (positive and negative), so we deal with 0.0 to 1.0 which is actually common practice in random number generation (and by extension security/cryptography).  

Here (and in most other terrain generation applications) would 0.0 be the equivalent to Mariana Trench and 1.0 would be Everest's summit.

https://en.wikipedia.org/wiki/Diamond-square_algorithm <-- easy to implement using positive real numbers 0.0 >= X >= 1.1

tcld:
I will remove this pull request as soon as possible. For now I think it is useful as a reference for me (and maybe for others to read).

tcld:
The CI build failed again due to a problem comparing floats. Maybe this is a case for assertAlmostEqual() ?

ftomassetti:
From what I recall there is one value passed to plate-tectonics and it is the value it attributes to sea-level while performing the simulation. I think that the sea-level affects the erosion calculated during plate-tectonics and how the clashes between plates are handled.
Then there is the value we used during normalization: we cover with water the required percentage of the world. Note that we start from the lowest point in the world and we grow the oceans until we have covered enough space. This approach permits to have depressions (land with elevation inferior to the sea level). During the plate-tectonics simulation instead what is below the sea level is just ocean.
To get a more precise answer I think we should dig into the plate-tectonics code.

About breaking the deterministic result: that is ok, some changes affect the final result. In these case I look at the result, decide if it is good enough and use it as a new reference. Depending on the tests that is broken you could have to generate new "blessed images" (see https://github.com/Mindwerks/worldengine/blob/master/tests/blessed_images/generated_blessed_images.py)

tcld:
The deterministicness is broken in a bad way: When using the same seed twice, the results differ wildly. I am currently trying to trace the seed-value and see where this might happen. Should I figure out the problem, I will update this patch and reopen it.

psi29a:
I would love to see 'blessed images'  be smaller in size... because it takes so long on Travis to test.  If that is undesirable, we can just not run that particular test during CI and only when manually running the test-suite.

ftomassetti:
We could reduce a bit the blessed images but the larger they are the most probable it is that we cover all the possible weird situations that can happen. Of course we could write more unit tests and reduce the need for this sort of functional test.

tcld:
I think this branch is almost finished. Here some images - 95% ocean, 65% ocean and 30% ocean, in that order:
seed_7049_biome
seed_7049_biome
seed_7049_biome

The worlds all used the seed 7049 but they seem quite different - as far as I understand things this should be platecs doing (otherwise the geometry should be the same and the following steps would differ, right?).

The only thing that is missing is a small fix to one of the test-files (comparing serialized and original ocean_level float currently gives an error due to floating point imprecisions). Since I for sure modified that file in the perf_opt-branch I don't know if I can make changes to it here (I think git would consider the changes a conflict).

psi29a:
Why did you delete and recreate the branch?

You can just do what I suggested:
```
git rebase -i master
```

tcld:
At some point I did *git fetch upstream* to get the latest changes to master and I wasn't able to push the ocean-branch afterwards. At that point I thought that it might not be useful anyway and... well, I think it is too late for a rebase.

I did use your suggestion for the accidentally committed World_pb2.py, though. It worked and I will remember it for next time. Sorry if my commits are a bit messy at the moment, I only ever used git for private projects so there were a lot less cables to stumble over.

psi29a:
You're doing fine! :)

Great work!

PR #127, opened 2015-10-13
Seed fix

tcld:
A very small change. I found it very confusing that the parameter "seed=0" resulted in a random seed even though internally 0 is a perfectly valid seed value. Since negative parameters should be equally valid anyway, only cutting out 0 seemed like an oversight.

Also, I changed 65536 to 65535. randint() includes its boundaries, so the former number results in a range of 2^16+1 (and it really looked like it was meant to be 2^16). Extremely minor and unimportant, I guess. But I was already changing the code anyway...

### DIFF
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -389,7 +389,7 @@ def main():
             usage("The specified world file does not exist")
 
     random.seed()
-    if args.seed:
+    if args.seed is not None:
         seed = int(args.seed)
     else:
         seed = random.randint(0, 65536)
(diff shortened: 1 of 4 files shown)


ftomassetti:
Looks good to me. @tcld thank you, you are doing a lot of great work for this project!

tcld:
Well, this is a really small patch.
I compiled an ran it by now, generated some worlds - so far everything seemed to work.

ftomassetti:
Cool, I think we can merge this one. @psi29a let me know if it is a problem and I will revert it

ftomassetti:
By the way we should also add you to the contributors in the README. Feel free to open a PR with the desired changes (your name, and the area you worked on, as we did for others contributors)

tcld:
Should I ever feel that's necessary, I will do that. Thanks. :)

PR #129, opened 2015-10-13
Simplify requirements

psi29a:
throwing it against the wall to see what sticks.. ;)


tcld:
Adding
+protobuf==3.0.0a3
to requirements.txt will enforce that version for Python 2 and 3. Does that version exist for Python 2? Does it work correctly? I didn't mind splitting the requirements for the different Python versions.

The rest certainly looks alright, but the tests seemingly don't agree.

tcld:
Ok, the failed build requires a very easy fix - you didn't rename requirements2.txt somewhere.

psi29a:
protobuf 3.0.0a3 is compatible with python2 and 3,  protobuf2.6 was not. :)

tcld:
Ah, I see you found the appveyor problem. :)
Just for completeness, you might want to change line 6 in setup.py, too. It's not functional, but some day it will look strange.

Great to hear that protobuf is available for both Python 2 and 3. So if the interface between 2.6.0 and 3.0.0 changed in any way, worldengine can easily be updated.

tcld:
By the way: .travis.yml and appveyor.yml contain specific instructions for only Python 2.7. Work for another commit? I don't know how to handle those files or I would have tried already.

psi29a:
And the % change in coverage is now heavily reduced! Great work @tcld 

Yes, you can be specify in travis and appveyor, but usually in scripts setting global variables and other stuff. It can get hairy and I don't like it. ;)

Set up your own branch like I mentioned before and just test! Nothing here bites hard! :P

tcld:
I have been doing that all day. :P 
Right now I have to come back to the Python 3 fixes from earlier, hopefully finishing them so that I can compile worldengine again.

tcld:
I tested the World_pb2.py:
My tests do not complain about unicode() anymore (supposedly 36 tests pass, 4 fail - probably related to pickle, same stuff as before). Compiling and running seems to be fine, too. :)

psi29a:
Super!  Same thing I get.

psi29a:
Merged, gets us into a better state. Stil failing python3 tests, but we're getting there.

tcld:
The rest of the problems *might* be related to pickle. Some are for sure, so maybe a switch wouldn't hurt right now.

tcld:
I just happened to run another test with your World_pb2.py, this time non-obstructed by my own code:
Only two tests failed (38 passed). I copied the output to pastebin:
http://pastebin.com/d5X529dM
It is possible that the second failed test has to do with the first one, or maybe the blessed images need to be refreshed since the random generator is now seeded differently. I don't really know what to do about either of these tests - but things look promising. :)

It would be great to get another release going at some point, it can be a little difficult to get worldengine running.

psi29a:
No worries, it is what I get too with python3.

python2 on the other hand works without a problem.

tcld:
I am actually just curious if the last two errors could be resolved by leaving pickle behind. I may not know what the tests are actually doing and what passing all tests means, but it is always nice to pass all the tests.^^

psi29a:
I'm thinking that there is some type of binary incompatiabilty, for instance that in python3 all strings are unicode while in python2 it can be either ascii or unicode. Another is int and long in python2 while in python3 long no longer exists as all ints are now longs by default.

tcld:
Yes, that is what the first failed test is about. And it is inside from_pickle_file(cls, filename) - so switching to protobuf would certainly circumvent that problem. After that the blessed images could be regenerated, maybe the last failed test solves itself.

PR #132, opened 2015-10-14
A bit of NumPy

tcld:
I played around with a profiler, picked out one of the slowest functions and took a look at what could be improved - I found a couple of smaller things, but nothing major. I then proceeded to see what numpy could do - but it actually doesn't do much. Speed improvements are within the margin of error, memory improvements weren't measurable due to me not knowing how to properly use memory_profiler.

The generated maps are 100% the same as before, result of the tests hasn't changed either.

Here some more information, not very useful and quite boring: http://pastebin.com/8p1vwaV1

### DIFF
--- a/worldengine/simulations/irrigation.py
+++ b/worldengine/simulations/irrigation.py
@@ -10,6 +10,11 @@ class IrrigationSimulation(object):
 
     @staticmethod
     def _calculate(world):
+        #Notes on performance:
+        #  -method is run once per generation
+        #  -iterations        : width * height * (2 * radius + 1)^2
+        #  -memory consumption: width * height * 8 Byte (numpy 1.9.2)
+
         width = world.width
         height = world.height
         radius = 10
--- a/worldengine/simulations/irrigation.py
+++ b/worldengine/simulations/irrigation.py
@@ -59,15 +59,17 @@ class IrrigationSimulation(object):
 
         values.transpose()
 
+#        values = [[0 for x in range(width)] for y in range(height)]  # TODO: replace with numpy
+#
 #        for y in range(height):
 #            for x in range(width):
 #                if world.is_land((x, y)):
 #                    for dy in range(-radius, radius + 1):   # TODO: below can be simplified
-#                        if (y + dy) >= 0 and (y + dy) < height:
+#                        if (y + dy) >= 0 and (y + dy) < world.height:
 #                            for dx in range(-radius, radius + 1):
-#                                if (x + dx) >= 0 and (x + dx) < width:
+#                                if (x + dx) >= 0 and (x + dx) < world.width:
 #                                    dist = math.sqrt(dx ** 2 + dy ** 2)
-#                                    values[y + dy, x + dx] += \
+#                                    values[y + dy][x + dx] += \
 #                                        world.watermap['data'][y][x] / (
 #                                            math.log(dist + 1) + 1)
 
--- a/worldengine/simulations/irrigation.py
+++ b/worldengine/simulations/irrigation.py
@@ -45,7 +45,7 @@ class IrrigationSimulation(object):
                                   x + it_logs.multi_index[1] - radius)
 
                     if 0 <= pt[0] < height and 0 <= pt[1] < width:
-                        values[pt] += world.watermap['data'][y][x] /logs[it_logs.multi_index]
+                        values[pt] += world.watermap['data'][y][x] / logs[it_logs.multi_index]
 
                     it_logs.iternext()
             it_all.iternext()
--- a/worldengine/simulations/irrigation.py
+++ b/worldengine/simulations/irrigation.py
@@ -1,5 +1,4 @@
 import numpy
-#import math
 
 class IrrigationSimulation(object):
     @staticmethod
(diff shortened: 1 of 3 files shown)


tcld:
Sh**, I accidentally added the World_pb2.py I got from @psi29a . Could somebody tell me how I can undo that?

tcld:
Oh, the build failed. When I return that numpy-array for the irrigation, it causes problems whenever something like "world1 == world2" is called since that array seemingly differs from a standard array more than expected.

I will think about it.

tcld:
I pushed another commit, now it hopefully works. I tried to test the addition to the protobuf-code (in world.py) but I am not sure what triggers it. "worldengine --protocol-buffer" ran fine.

I also did some testing regarding the memory consumption of a numpy array vs. a standard array. Outputs are read off of the Ubuntu Task Manager. Imprecise for small arrays. And for some reason there are two values - private and (virtual). The number on the left is the size of the array (e.g. 1000*1000).

old:
     1000^2: 5 (40) MB ->   14 (  48) MB
     4000^2: 5 (40) MB ->  135 ( 169) MB
    10000^2: 5 (40) MB ->  808 ( 876) MB
    20000^2: 5 (40) MB -> 3134 (3466) MB

new:
     1000^2: 5 (40) MB ->   17 (  93) MB
     4000^2: 5 (40) MB ->   17 ( 207) MB
    10000^2: 5 (40) MB ->   17 ( 848) MB
    20000^2: 5 (40) MB ->   17 (3137) MB

Creation and destruction are *a lot* faster for the numpy arrays. Memory consumption in comparison to standard arrays is slightly better for large arrays. As far as I understand it's the iteration that isn't much faster - but there are some optimized iteration possibilities, maybe they prove useful in other places of worldengine.
It was certainly worth a try, and depending on what you guys think about this and if the tests will start going through at some point, I might start looking at some of the other code-files and add in some numpy-code - hints on where it might be most profitable are welcome. (I don't have to go all out as with this commit, I just wanted to give it a fair shot.)

tcld:
Yippie, the tests are passed! :) (Except for the lottery-test...)

psi29a:
Nice work! We've had this on our todo list for awhile. :)  

I think there are more places in the code where the 2d arrays should be numpy.

tcld:
Yes, there are. It would be great if all of them were, that might eliminate some overhead converting between the two types and give a very clean memory-footprint and small speed improvements.

When I started editing the function I just wanted to exchange the array for a numpy array, due to the TODO-remark. When I then dug into the documentation I noticed that there is a lot more that could be done with numpy - sadly the improvements are pretty minor, I was hoping for at least a factor of two in speed.

tcld:
I'm still polishing this, so please nobody merge it. numpy is kind of insane. :)

psi29a:
I'm a little worried about your revert of World_pb2.py, since what was in master was compatible with python2 and python3.

Did it not work with your version of protobuf?

tcld:
Oh, no! It worked perfectly. I guess I used "commit -a" when I should have used "commit". I'm sorry. :(
I will remove it. Again.

psi29a:
No worries, nothing that can't be undone. I was just reading through your change-set and noticed it. :) 

When you are ready for us to really look at it (and merge), just say so. We'll likely have you do a:
```
rebase master -i
```
which will allow you to squash all your commits into one. This makes your commit changes super clean as a result. :)

tcld:
Thanks for the hint. :)

The code as is is in a state of readiness. But numpy seems to be very powerful and I am currently trying to find a way to get rid of most of the iterations in the modified file. It would be nice to have a really polished example of numpy-usage in the code, maybe somebody else decides to put more numpy into worldengine and needs a reference. I, for one, had no idea that numpy would prove to be such a deep pond.

(It's a shame I will have to throw away my coverage-lottery ticket with the next commit, though.^^)

tcld:
I might be done. Offline-tests ran fine, result of the generation is exactly the same as with the old algorithm and the code now looks quite beautiful to me.^^
I desperately have to run some performance tests on this because this now needs two orders of magnitude less iterations and the things that are done should be doable very efficiently. The old "new" method saved about 10% of time, hopefully the current one is at least as good.

tcld:
Yippie, cProfile reports about 90% less time spent in irrigation.py - which is about 25% less time overall! :dancers:

psi29a:
Python unlike C/C++ (compiled languages) is an interpreted language. It will write bytecode out when the file is first run into *.pyc files. Python is smart enough to invalidate pyc files when the py files are changed but will only actually run the pyc files.

So I'm not really sure what you mean by compiled master-build and uncompiled test-build.

If you mean our one file things found in our github releases page, that is a whole other ball of yarn having to do with `pyinstaller` black-magic. That assembles all the different libraries and Python bytecode into one file that resembles a self-extracting zip file.

tcld:
I meant: I properly installed the master-build. Then uninstalled, removed all traces of .pyc (I don't trust Python because of some mishap in the past) and *then* reran the script.

But it might not be important if there is any more time that can be shaved off of the 0.2 seconds that are left. :P

psi29a:
For sake of nailing down definitions... let us just call it based on the branch.  
How about just master vs. perf_opt
or if you are talking about a release, use the tag name like  0.18 vs. perf_opt 
it is safe to say your branch or tag vs. another branch and tag because they are unique. :)

tcld:
That seems more clear, I will remember it.^^

Meanwhile I am looking through the profiler-output to find some more timesinks. There are some that might be worth looking into, but nothing really prominent.

This one is interesting, though:
```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
1000    8.378    0.008    8.378    0.008 {built-in method step}
```
Is that what is done when iterating through a range? I have never seen a step()-method - and a quick look into the Python documentation didn't help either.

worldengine/simulations/irrigation.py
```diff
@@ -11,22 +10,42 @@ def execute(self, world, seed):
 
     @staticmethod
     def _calculate(world):
+        #Notes on performance:
+        #  -method is run once per generation
+        #  -iterations        : width * height
+        #  -memory consumption: width * height * 8 Byte (numpy 1.9.2)
+
```
psi29a: This is nice, I love documentation and I love precise memory allocation prediction and measurements. :)

psi29a:
Looks great, merging.

PR #137, opened 2015-10-16
Adopt pyplatec 1.4.0

ftomassetti:
Let's take advantage of the speed improvements we got in the new version of plate-tectonics/pyplatec

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,6 +2,6 @@ argparse==1.2.1
 noise==1.2.2
 numpy==1.9.2
 Pillow==2.8.2
-PyPlatec==1.3.1.post1
+PyPlatec==1.4.0
 protobuf==3.0.0a3
 six==1.10.0
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ config = {
     'entry_points': {
         'console_scripts': ['worldengine=worldengine.cli.main:main'],
     },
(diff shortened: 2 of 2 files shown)


tcld:
This is really, really great! When I last checked, about 25% of the overall time needed to generate a world went into platec. That could go down to a negligible amount.
I am not sure how my perf_opt-branch (https://github.com/Mindwerks/worldengine/pull/132) does with large worlds, but for smaller ones it already decreased generation times by about 25%. So overall a speed-up of up to a factor of 2 would be feasible for the "next version".

psi29a:
Looking good, merging!

@ftomassetti have you been in contact with our Finnish counterpart? We going to be up-streaming our patches?

ftomassetti:
Nope, I have not being in contact since he said he was leaving for holidays. I think our fork diverged quite a bit, I hope they are fine accepting our patches. Perhaps it is time to write again to them to present the significantly speedup we got (admittedly, it is possible that I slow down our code while refactoring and this speedup just makes as fast as platec was already originally :D, we should measure that)

psi29a:
Check your email, he sent something awhile back with you in CC. I'll forward it to you.

psi29a:
@ftomassetti I forwarded you the email he sent (with you in CC) in Sept. So he is back and has a plan. I also sent a reply. You sure you didn't get that?

Issue #136, opened 2015-10-16
Very Short Agenda


tcld:
Here is a short list of things that are currently on my mind. I have at least a couple more days that I could use to work on worldengine, let's see what I can get finished.
If there is anything here that you don't like, please tell me and I will focus on the other points first. If there is anything much more pressing, maybe I can look at that instead. But since my background-knowledge is at a flat zero, I cannot check or improve any of the generation algorithms. Optimizing is in the cards, though, thanks to numpy.

- [x] Replace several, maybe all, arrays with numpy arrays.
  * Top on the list: The ocean-map. Used extremely often, maybe a bit of speed can be gained.
  * elevation and watermap would be the next candidates.
- [x] Create a custom, recursive equality operator (in common.py?). Good for numpy and floats, at least.
- [ ] ~~Think about the funny rivers and try to fix them.~~ ( @psi29a called dibs)
- [x] ~~Look at the memory usage of worldengine. Maybe it is just platec, but something needs a *lot* of RAM.~~ (I didn't find any glaring holes.)
- [x] Add support for a different image format so heightmaps can store more than 256 heights.
- [x] Look at temperatures.py and give it an additional parameter to...do something nice :)

I will edit the list as I go.

ftomassetti:
Consider that we are going to release soon the new version of plate-tectonics (our platec fork) which is up to 6 times faster. Perhaps it also uses less RAM

psi29a:
Numpy is the way to go for ALL arrays. It has built in functions like find Max and Min that are super fast, we should be abusing these functions.

psi29a:
If you don't mind, I want to work on Rivers... I just need to find time. If you beat me too it, fair is fair. ;)

tcld:
Ok, rivers are yours then.^^ (but please specify which source files you mean, as there are two...kind of)

Abusing numpy is a good idea, but there is a lot more necessary than just replacing arrays. Many portions of the code would have to be rethought to make really good use of numpy. Iterating has to be avoided wherever possible, as can be seen here https://github.com/Mindwerks/worldengine/pull/132 . The very first iteration used a numpy-array, no further changes. That gives a nicer memory footprint (not even that much smaller) but that's it. The second iteration was basically an emulation of the former version that didn't use numpy. That sped things up by only 10%. Only once I went all the way and completely rethought things, things sped up (by a factor of 50!).

[...]

PR #138, opened 2015-10-16
Make use of Numpy-Arrays

tcld:
Another branch, as promised (https://github.com/Mindwerks/worldengine/issues/136). I replaced almost all arrays with numpy-arrays. Everything that is *not* mentioned in the serialization-tests is probably not touched, either. (I had to start somewhere.)

Here some performance-test results:

Parameters: 
```
  width        = 384
  height       = 256
  plates       =   5
  oceansurface = 65% (only effective in the numpy_repl-branch, which is based on ocean_level_fix)
  seed         = 13047
```

Generation times for every branch:
```
  master     = 39.57 s
  perf_opt   = 29.22 s
  numpy_repl = 28.40 s
  (numpy_repl + platec1.4.0 = 21.48 s)
  (numpy_repl + platec1.4.0+ small improvements = 20.44 s)
```

Images (master - perf_opt - numpy_repl - platec1.4.0):
seed_13047_biome
seed_13047_biome
seed_13047_biome
seed_13047_biome

The latest changes were not supposed to speed things up too much, only some of the excessively used functions were replaced. I did what I could while keeping an eye on the target of getting numpy ready. In the future it won't be a pain to get a bit of optimization done anymore. And maybe the memory footprint will have changed a bit, I still have no way of reliably checking that.
I will do some tests with larger maps and see if anything changed there. (And I am really looking forward to the new version of platec. :) )

All tests except two are passed, although I did run one of the regeneration-scripts in the tests-folder to make it happen. At this point, especially due to the change to the seeding, worldengine-data should be updated a bit anyway, I think. (One of the failed tests, maybe even both, can probably be passed with fresh data - maybe it has to be Python 2/3-specific, though.)

EDIT: I took a look at the new platec version and added the time above. In total we are down to roughly 50% of the generation time from a couple of days ago! :)

EDIT2: As far as I see, the failed test comes from the fact that the old world-file was generated without numpy-arrays. I wouldn't know what to do about that - on my system regenerating the test-world was enough, though.

EDIT3: I did a little more profiling this evening. The air gets thin now, there are only three methods left that slightly justify spending time on with optimization (world.is_land()~11% of total time, world.tiles_around()~20%, common.anti_alias_point()~20% - that's roughly 50% in total!). And after looking at them for a while and trying some things I really couldn't come up with any improvement.
A funny sidenote is that the function calls for my testmap went done from more than 300 million to less than 100 million. Half of which is attributed to is_land(). Speed-wise this might be the end of the road, unless a lot more work is invested.

```
         95833445 function calls (95705636 primitive calls) in 92.104 seconds

   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  4014080   19.543    0.000   19.543    0.000 worldengine/../worldengine/common.py:97(anti_alias_point)
  1262482   18.308    0.000   27.963    0.000 worldengine/../worldengine/world.py:375(tiles_around)
     1150   14.349    0.012   14.349    0.012 {built-in method step} -- this refers to platecs step()-method
 56451705   10.417    0.000   10.417    0.000 worldengine/../worldengine/world.py:341(is_land)
        1    2.312    2.312    4.155    4.155 worldengine/../worldengine/simulations/erosion.py:136(river_sources)
```


psi29a:
Doing a rebuilt based on new master.

tcld:
As said here https://github.com/Mindwerks/worldengine/pull/134 , I think that one of those last two commits (well, I hope it is one of those) might have introduced a problem. Once the results of a currently running generation are done, I will post an image, should anything go wrong. (Or just a message, if everything is ok.)

tcld:
Update on this:
I separated this branch from the ocean_level_fix-branch (https://github.com/Mindwerks/worldengine/pull/134). This branch passed all minus two tests for me (after running tests/data/data_generator.py as far as I remember) and it still produces exactly the same output as the master-branch. In about 75% of the time, I have to add. ;)

worldengine/simulations/humidity.py
```diff
@@ -17,12 +18,9 @@ def _calculate(world):
...
         precipitationWeight = 1.0
         irrigationWeight = 3
-        humidity['data'] = [[0 for x in range(world.width)] for y in
-                            range(world.height)]  # TODO: replace with numpy
+        humidity['data'] = numpy.zeros((world.height, world.width), dtype=float)
 
-        for y in range(world.height):
-            for x in range(world.width):
-                humidity['data'][y][x] = (((world.precipitation['data'][y][x] * precipitationWeight - world.irrigation[y][x] * irrigationWeight)+1) * ((world.temperature_at((x, y)) * temperatureWeight + 1.0)/(temperatureWeight + 1.0)))
+        humidity['data'] = (((world.precipitation['data'] * precipitationWeight - world.irrigation * irrigationWeight) + 1) * ((world.temperature['data'] * temperatureWeight + 1) / (temperatureWeight + 1)))
```
tcld: numpy is just awesome. :D

worldengine/generation.py
```diff
@@ -144,26 +140,24 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
...
     sea_depth = anti_alias(sea_depth, 10)
-    min_depth, max_depth = matrix_min_and_max(sea_depth)
-    sea_depth = [[rescale_value(sea_depth[y][x], min_depth,
-                                max_depth, 0.0, 1.0)
-                  for x in range(world.width)] for y in
-                 range(world.height)]
+
+    min_depth = sea_depth.min()
+    max_depth = sea_depth.max()
+    sea_depth = (sea_depth - min_depth) / (max_depth - min_depth)
```
tcld: To be able to use numpy I copied the functionality from here:
https://github.com/Mindwerks/worldengine/blob/master/worldengine/common.py#L74

tests/serialization_test.py
```diff
@@ -1,16 +1,9 @@
...
 import tempfile
 import os
-import numpy
-
-
-def _sort(l):
-    l2 = l
-    l2.sort()
-    return l2
-
```
tcld: I think this was redundant. Pythons sorted()-method does the job.

worldengine/astar.py
```diff
@@ -158,11 +158,10 @@ def __init__(self, map_data, width, height):
     def get_node(self, location):
         x = location.x
         y = location.y
-        if x < 0 or x >= self.w or y < 0 or y >= self.h:
-            return None
-        d = self.m[(y * self.w) + x]
-
-        return Node(location, d, ((y * self.w) + x))
+        if (0 <= x < self.w) and (0 <= y < self.h):
```
tcld: I often read (never checked, though) that (a < x < b) is a lot faster than (a < x and x < b).

worldengine/astar.py
```diff
@@ -198,14 +197,6 @@ def _handle_node(self, x, y, from_node, destination_x, destination_y):
         return None
 
 
-def _matrix_to_array(matrix):
```
tcld: This method ate huge amounts of time in comparison to everything else. I hope the replacement below is ok, but I am scratching my head about that a bit. (tests and output, however, are perfect)

worldengine/common.py
```diff
@@ -80,32 +81,31 @@ def rescale_value(original, prev_min, prev_max, min, max):
     return min + ((max - min) * f)
 
 
-def anti_alias(elevation, steps):
+def anti_alias(map, steps):#TODO: There is probably a bit of numpy-optimization that can be done here.
```
tcld: I spent an hour trying to optimize this, making heavy use of numpy. For some reason that ended up even slower, though. Somebody else could try himself, since this method is one of the last few that overall take a huge amount of processing time (20% of the toal).

worldengine/common.py
```diff
@@ -120,3 +120,33 @@ def array_to_matrix(array, width, height):
         for x in range(width):
             matrix[y].append(array[y * width + x])
     return matrix
+
+def _equal(a, b):
```
tcld: I hope this doesn't seem too much like a hack. I just didn't know how to handle numpy.array== otherwise (it defaults to returning an array of bools instead of a single bool, which did cause problems at many places in the code).

worldengine/draw.py
```diff
@@ -176,9 +177,9 @@ def draw_riversmap(world, target):
     sea_color = (255, 255, 255, 255)
     land_color = (0, 0, 0, 255)
 
-    for y in range(world.height):
+    for y in range(world.height):#TODO: numpy
         for x in range(world.width):
-            if world.ocean[y][x]:
+            if world.ocean[y, x]:
```
tcld: The meat of this branch: Switching out [y][x] for [y, x]
I actually tested this and yes, the latter version is a bit faster. And with so many arrays in memory and such an incredible amount of calls to the operator[] I thought it might be worth doing this replacement to gain a little bit of speed overall (it amounted to ~1-2% in total).

worldengine/world.py
```diff
@@ -94,7 +85,16 @@ def _to_protobuf_matrix(matrix, p_matrix, transformation=None):
         for row in matrix:
             p_row = p_matrix.rows.add()
             for cell in row:
-                value = cell
+                '''
```
tcld: This, too, is a little bit dirty and made necessary because of the numpy-arrays.  The only bool-map I know of is the ocean-map, but that one is used so frequently that I didn't want to leave it without numpy-support.

tcld:
I added a ton of comments since these commits change quite a bit. Hopefully I didn't make anything unnecessarily cryptic.

psi29a:
```
  File "/home/travis/build/Mindwerks/worldengine/worldengine/drawing_functions.py", line 101, in _find_land_borders
    if world.ocean[int(y / factor), int(x / factor)]:
TypeError: list indices must be integers, not tuple
```

This looks to me that ocean isn't numpy...

psi29a:
```
  File "/home/travis/build/Mindwerks/worldengine/worldengine/world.py", line 483, in elevation_at
    return self.elevation['data'].T[pos]
AttributeError: 'list' object has no attribute 'T'
```

Is this a typo?

tcld:
No. "T" is a call to the transpose-function of the numpy array.
http://docs.scipy.org/doc/numpy/reference/generated/numpy.ndarray.T.html

The error only occurs when an old world is used. However, I can replace the call to T for something more clear, it doesn't gain any performance anyway. (That's why I tried it.)

tcld:
It is "fixed" now. I should have done that earlier.

tcld:
I didn't notice this before:

> This looks to me that ocean isn't numpy... 

I have not really understood why but that is the same problem as with the missing transpose() method: Somehow when loading an old map the arrays are not (yet?) numpy arrays. Did I overlook a place where the arrays could get into the program without being converted? Is this due to the pickle-format?
After regenerating the test-world this worked on my local system. Not understanding why, though, is a problem, of course.

tcld:
I think the problems come from this:
https://docs.python.org/2/library/pickle.html#pickle.load

I assume the test-world is a pickle-world and things go like this:
1. https://github.com/tcld/worldengine/blob/numpy_repl/worldengine/cli/main.py#L481
2. https://github.com/tcld/worldengine/blob/numpy_repl/worldengine/cli/main.py#L208
3. https://github.com/tcld/worldengine/blob/numpy_repl/worldengine/world.py#L50

The Protobuf-loaders could be manipulated in World.py, but the old pickle-save does not contain numpy-arrays and thus after loading the arrays cannot be used as such.

If this sound feasible, we should really regenerate that world and see how the two tests do that so far have been failing (for me, at least).

psi29a:
That was my first thought as well... loading old datafiles will have data as array and not numpy array.

This means we'll likely have to regenerate the world files to accept the new format. :)

We'll be breaking compatibility with older world files as a result.

tcld:
All the arrays are converted after being loaded respectively before being saved with protobuf. I expect old protobuf saves to still work; I don't have one to test it, though.

I never made the biome-map use protobuf (but I think that is the only one, it just looked like quite a bit of work). So if the world were to be regenerated...at some point another break of compatibility would have to occur, I fear.

tcld:
I think I might at least put very basic code in there to have the biome-map be a numpy-array (right now), without the usual small optimisations that follow all over the code. That way we don't have to break pickle-compatibility twice. (By the way: If protobuf were to be the favourite serializer, maybe the test-worldfile should be in protobuf-format? That way we wouldn't need two separate files for Python 2 and 3, either.)

Or is this kind of break maybe too much of a problem and the switch to numpy not feasible? I would hate to see that.

psi29a:
I agree totally... 

@ftomassetti I think we're getting to that point where we need to pull the trigger on axing pickle. ;)

psi29a:
In worldsynth, I used pytables which created a sort of SQL layer above HDF5. It supported inline compression and other nifty stuff. This is always an option btw...

tcld:
Sounds great as an option for the future. :D Right now I want to get my branches compatible. I continue building on top of them and that feels like a bad idea when the base isn't 100% solid yet.

psi29a:
Again, I agree. :)

tcld:
This went surprisingly quickly. This branch should now replace all major arrays with numpy arrays. The next pull request https://github.com/Mindwerks/worldengine/pull/141 will take care of smaller ones that are spread around the code. To pass the tests it will still need a freshly generated world (generated with this branch, right? So maybe I should do that?), but it passed all tests that I threw in its way.

ftomassetti:
Dropping pickle is fine for me: I just started using it because it was the easiest thing to use in the beginning. I am all in favor of formats which are easily reusable from other languages, that was my goal with protobuf. If protobuf does not convince you I am absolutely open to alternatives.

psi29a:
Before we merge this in, can you try fixing the failing test?

Fork https://github.com/Mindwerks/worldengine-data and send a PR with a world file using the same seed number.

Once this is merged, we can retry the build here and merge. :)

tcld:
So...I fork worldengine-data, add a freshly generated world (Using worldengine/tests/data/data_generator.py ?)

There are three worlds:
https://github.com/tcld/worldengine-data/tree/master/tests/data

Why are there three? Which one will I replace?

worldengine/astar.py
psi29a: For your edification and ours, can you profile this after 100,000 iterations?

The original takes advantage of 'or' short-cutting so it doesn't have to evaluate the rest. The one on the right has to evaluate the whole thing.

I also believe that your parenthesis are redundant.

worldengine/astar.py
psi29a: Perfect, that is numpy's strong point. good job!

worldengine/biome.py
```diff
@@ -219,6 +219,6 @@ def biome_name_to_index(biome_name):
 
 def biome_index_to_name(biome_index):
     names = sorted(_BiomeMetaclass.biomes.keys())
-    if biome_index < 0 or biome_index >= len(names):
+    if not (0 <= biome_index < len(names)):
```
psi29a: Are those parans necessary?

worldengine/cli/main.py
```diff
@@ -95,7 +96,7 @@ def generate_plates(seed, world_name, output_dir, width, height,
                                                    num_plates=num_plates)
 
     world = World(world_name, width, height, seed, num_plates, -1.0, "plates")
-    world.set_elevation(array_to_matrix(elevation, width, height), None)
+    world.set_elevation(numpy.array(array_to_matrix(elevation, width, height)), None)
     world.set_plates(array_to_matrix(plates, width, height))
```
psi29a: Since you flattened 2d array into 1d array with numpy, I know that numpy can take a 1d array and `numpy.reshape` into an Xd array.  Can you replace this function too please? :)

worldengine/biome.py
tcld: I don't know. What do you say?

worldengine/common.py
psi29a: Is this function used in other parts of WE or just tests? If just tests, can it be moved to the tests please?

worldengine/simulations/basic.py
```diff
@@ -1,14 +1,15 @@
-def find_threshold(elevation, land_percentage, ocean=None):
-    width = len(elevation[0])
-    height = len(elevation)
+import numpy
+
+def find_threshold(map, land_percentage, ocean=None):#never used anywhere?
+    height, width = map.shape
```
psi29a: can we get away from using `map`? rename back to elevation or use something generic like map_data or data?

map is a keyword: https://docs.python.org/2/library/functions.html#map

worldengine/common.py
tcld: It is used in/as World.operator== - removing it would leave worlds without a proper way of checking for equality. I said this already at some point, but: I think this method should stay here.

worldengine/simulations/basic.py
```diff
@@ -33,36 +34,32 @@ def search(a, b, desired):
...
     desired_land = all_land * land_percentage
     return search(0, 255, desired_land)
 
 
-def find_threshold_f(elevation, land_perc, ocean=None):
-    width = len(elevation[0])
-    height = len(elevation)
-    if ocean:
-        if (width != len(ocean[0])) or (height != len(ocean)):
+def find_threshold_f(map, land_perc, ocean=None, max=1000.0, mindist=0.005):
```
psi29a: can we get away from using map? rename back to elevation or use something generic like map_data or data?

map is a keyword: https://docs.python.org/2/library/functions.html#map

worldengine/simulations/humidity.py
psi29a: agreed :)

worldengine/world.py
```diff
@@ -416,22 +415,10 @@ def start_mountain_th(self):
         return self.elevation['thresholds'][2][1]
 
     def max_elevation(self):
```
psi29a: replace all calls to max_elevation to the .max(), saves us time calling a function just to get to numpy's builtin.

worldengine/world.py
```diff
@@ -416,22 +415,10 @@ def start_mountain_th(self):
...
 
     def min_elevation(self):
-        min_el = None
-        for y in range(self.height):
-            for x in range(self.width):
-                el = self.elevation['data'][y][x]
-                if min_el is None or el < min_el:
-                    min_el = el
-        return min_el
+        return self.elevation['data'].min()
```
psi29a: replace all calls to min_elevation to the .min(), saves us time calling a function just to get to numpy's builtin.

worldengine/biome.py
psi29a: If they are not necessary, ditch them please. :)

psi29a:
I don't know, that is why I asked to replace the one with the failing seed... you can specify the seed number running WE manually. The resulting world file should be good enough right?

```
    def test_draw_ancient_map_factor1(self):
        w_large = World.from_pickle_file("%s/seed_48956.world" % self.tests_data_dir)
        target = PixelCollector(w_large.width, w_large.height)
        draw_ancientmap(w_large, target, resize_factor=1)
        self._assert_img_equal("ancientmap_48956", target)
```

That would be the seed_48956.world (just use 48956 as your seed), make sure to ask for an ancient_map as output. Commit those two, and that should fix one test. :)

tcld:
Which parameters do I use for that world?

psi29a:
Create a new one, just specify that seed. 

You can then run we again, this time with ancientmap option and use the resulting world file from above.

tcld:
I won't touch the images at all, none of them is named in a way that makes me sure if I should replace it or not. If they were supposed to, they would have their standard names so I could easily overwrite them...right?

worldengine/astar.py
tcld: I used these
```
timeit.timeit('0 <= 50 < 500 and 0 <= 301 < 300', number=10000000)
timeit.timeit('50 < 0 or 50 > 500 or 301 < 0 or 301 > 300', number=10000000)
```

to profile this. The latter method seems to be faster: 0.907968218998576 vs 0.8031158160010818
So I reverted this piece of code. Pushing after your other comments have been worked on.

worldengine/astar.py
tcld: I just profiled a<b<c to gain about a millisecond while this change gained about 25% of the total runtime. :D I shouldn't try to look into those super-small things anymore. :P

worldengine/biome.py
tcld: I ditched them after a test. They were not necessary.

worldengine/simulations/basic.py
tcld: I forgot about that, thank you. Changed now.

worldengine/world.py
tcld: Then remove this function or leave it there for "emergencies"?

worldengine/world.py
tcld: Ok, I decided to remove these.

tcld:
Of course I touched all the images. And I think I updated everything that was possible - pickle-save (and the one corresponding image) excluded.
https://github.com/Mindwerks/worldengine-data/pull/1

worldengine/cli/main.py
tcld: You are right, it can do that. Is elevation a 1d-array, though? I didn't notice. Maybe it's not time to sleep yet, after all.

tcld: It's done. Thanks for the hint, I totally missed this.

tcld:
Another commit. Now the only thing that is open to debate is this:
https://github.com/Mindwerks/worldengine/pull/138/files#diff-6370c3dd8fe78487da3ac1540b8d3b4cR124

I would suggest leaving it in. There are quite a lot of actually dead functions (maybe a cleaning pass should be underdone), this one at least provides the World class with a working equality operator.

tcld:
I could use some help right now: I generated a full set of data, including that for Python 2. Under Python 3 all tests pass, under Python 2 the ancient-map tests don't pass (that is 2/40). Did I do something wrong while generating?

tcld:
Here the coloured version - Python 3 first, Python 2 last:

ancientmap_28070_factor3ancientmap_28070_factor3
The mountains in the middle are oriented differently.

ancientmap_48956ancientmap_48956
These two are *very* different. :(

tcld:
I will take another look at this tomorrow morning. Maybe I made some mistake that happens to make Python 2 and 3 create different outputs.

tcld:
I think I found the problem.

```
Python 3.4.3 (default, Oct 14 2015, 20:28:29) 
[GCC 4.8.4] on linux
>>> import random
>>> random.seed(0)
>>> random.randint(0, 1000000)
885440
```
```
Python 2.7.6 (default, Jun 22 2015, 17:58:13) 
[GCC 4.8.2] on linux2
>>> import random
>>> random.seed(0)
>>> random.randint(0, 1000000)
844422
```

Maybe we need a new random module. And a new noise module. :(

For now I guess I will push all my changes and then we regard this is a problem to be fixed.

ftomassetti:
I am wondering: did you generate the two random numbers on the same machine just using different version of Python? I ask because I read "linux" and "linux2". If the random generator gives different numbers on different versions of Python that is bad :(
I am looking for library to replace it. I know it is frustrating to fix those tests but hopefully they will protect us by unintended changes.

tcld:
Just opened an issue: https://github.com/Mindwerks/worldengine/issues/145
Yes, same machine. But as can be seen different compilers.

The changes in the Python-module can probably be spotted and compared. I don't know about gcc, though.

Should I go through with the pushes and we sort this out later?

tcld:
Finally: Here https://github.com/Mindwerks/worldengine-data/pull/2 is the PR for the actual, final, eternal data that makes *Python 2* pass all tests.

If that PR is merged and the tests are rerun, things should work fine. I really, really hope. :)

tcld:
How do you trigger a re-test?

tcld:
Yippie! :)

psi29a:
Make it so @ftomassetti ? :) I'm happy with this.

tcld:
What's the situation here?

ftomassetti:
Fine by me: all tests are passing :)
Should we hit the merge button or is there anything else any of you want to check?

tcld:
I didn't want to change anything about this branch anymore, no. The ones after might need some polish, though.

ftomassetti:
Merged, now we can see what adaptations the other PRs require

PR #141, opened 2015-10-18
Numpy+images

tcld:
This makes more use of numpy, even though only for the exports (and those don't take too much time anyway). I also put some work into the GDAL export() method; there might have been a bug or two and I wanted to give worldengine the opportunity to make use of more of the formats GDAL supports. (Even though I don't like GDAL since it was a pain to install. I created an instruction, though. Maybe it can be added somewhere.)

In draw.py I added an import for the PyPNG-library which is used to export 16 Bit grayscale images.

All of this is based on this https://github.com/Mindwerks/worldengine/pull/138 and there may be some things in here that need a bit of discussion. Due to some of the changes and the connection to https://github.com/Mindwerks/worldengine/pull/138 this will not pass the tests for now.

EDIT: Oh, Travis doesn't know the PyPNG-module I imported. Well, this still needs a tiny bit of polish. :)


worldengine/common.py
```diff
@@ -120,3 +120,33 @@ def array_to_matrix(array, width, height):
         for x in range(width):
             matrix[y].append(array[y * width + x])
     return matrix
+
+def _equal(a, b):
+    #recursion on subclasses of types: tuple, list, dict
+    #specifically checks             : float, ndarray
```
psi29a: _equal code doesn't seem to be used by the WE library. Can this be moved `__init__.py` in tests?

tcld: It is also used as World.operator== - I am not sure if there are any world-comparisons in the code outside of the tests. But I guess it is worth a try and the code can be moved, yes.

worldengine/imex/__init__.py
```diff
@@ -12,24 +12,49 @@
 import sys
 
 
-def export(world, export_type, export_bpp, export_signed, export_normalize):
+def export(world, export_filetype, export_bpp, export_signed, path = 'seed_output', export_raw = False):
```
psi29a: I don't like this, I expect export_raw by default when working with DEM. When you scale it, you're warping the original values. This is fine for visualizing in images, but otherwise very bad. You are destroying your data set to match your two extremes.

tcld: What is your suggestion? Remove the parameter or make it True by default?

tcld:
@psi29a If you want to look at the code, you should probably start with the branch this is based on (https://github.com/Mindwerks/worldengine/pull/138) unless you want to verify it all at once.

worldengine/common.py
tcld: On second thought the removal would leave the world-class without a functioning equality-operator. At the same time it is not only usable for that class... I wouldn't feel good about moving this to the tests, I prefer it here. (I found quite a lot of functions spread over various files that are never called or even referenced... maybe it is time for a bit of cleanup anyway.)

worldengine/imex/__init__.py
psi29a: Remove it? No way, you just renamed my export_normalize to export_raw. :)    Since we are giving the option to scale the data, thus normalize it, then the original name is more explanatory. That is only useful for image creation, which is why I had it in there to begin with.

I'm glad you are expanding it, I'm all for dropping Pillow and going full GDAL.

tcld: I did not rename your variable, I removed "export_normalize". When requesting integer output without scaling one would end up with numbers between 0 and maybe 15 - that kind of output is useless, especially when blown up to a 32 Bit image, so I enabled normalization by default. However, when requesting "raw" output (float32), normalization is detrimental at best, as you said.

Full GDAL is difficult in my opinion, I had a hard time installing it. PyPNG, however, is a nice middle ground. At least it is more capable than Pillow, although it is restricted to PNG. (Which I think is fine since that is a widely-spread, lossless format.)

psi29a:
You'll need to add PyPNG to tox.ini

tcld:
Ok, thanks. Since this https://github.com/Mindwerks/worldengine/pull/144#event-440746090 is now merged, I should probably rename the worldengine-data files to fit the new format (py3_seed...). I was away all day and didn't manage to do it before I left. I will finish everything up as soon as possible and hopefully the tests will run fine again. :)

tcld:
This is going to be broken down into pure numpy-rewrites of parts of the image-code https://github.com/Mindwerks/worldengine/pull/161, a rework of the GDAL export https://github.com/Mindwerks/worldengine/pull/159 and the 16 Bit grayscale heightmaps I promised somewhere (that branch is not yet fully separated; no PR for now).

PR #142, opened 2015-10-19
Two very small changes to the setup-files.

tcld:
Two tiny things that have made my eyes sore for a while. (Unimportant under almost all circumstances, yes.^^)

EDIT: And I think that we don't have to restrict ourselves to a specific version of numpy. I wouldn't expect major changes there. (Does that have to be changed in requirements.txt, too?)


psi29a:
:) Looks good!

PR #144, opened 2015-10-20
Tests: Differentiate between Python 2 and Python 3 saves

tcld:
If this is ok, I will follow up with an update to my PR at worldengine-data.

https://github.com/Mindwerks/worldengine-data/pull/1

### DIFF
--- a/tests/blessed_images/generated_blessed_images.py
+++ b/tests/blessed_images/generated_blessed_images.py
@@ -8,6 +8,7 @@ A script, generate_blessed_images, can be used to regenerate blessed images
 """
 
 import os
+import platform
 from worldengine.world import *
 from worldengine.draw import *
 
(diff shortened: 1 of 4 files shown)


tcld:
I remembered that World_pb2.py now supposedly is the same for Python 2 and 3. Is that right? And if so: Well...maybe we don't need separate pickle-files? Since I can't use the Python 2-version of Worldengine I also can't check what it would output and the files maybe are the same.

psi29a:
protobuf solves the python2/3 problem (one file fits all), but only for protobuf worlds. :)

tcld:
Oh, I'm an idiot. Disregard my last comment.^^ (I was thinking about pickle and somehow mixed the protobuf file in, and then thought "Hey, maybe the problem is non-existent?".)

PR #146, opened 2015-10-23
random -> numpy.random

tcld:
Here the replacement of all uses of the random-module with respective numpy-methods.
Since this is kind of a break with save-compatibility and...everything, I guess, I also took the freedom to include some small changes so they aren't necessary in the future - I will add comments to the code to point them out and explain them.

Notes:
1. This is based on https://github.com/Mindwerks/worldengine/pull/138 .
2. This does not pass the ancient-map tests (since that one uses RNG to determine its outcome)

### DIFF
--- a/tests/basic_map_operations_test.py
+++ b/tests/basic_map_operations_test.py
@@ -1,21 +1,12 @@
 import numpy
 import unittest
-from worldengine.basic_map_operations import distance, index_of_nearest, random_point
+from worldengine.basic_map_operations import distance, index_of_nearest
 
 
 class TestBasicMapOperations(unittest.TestCase):
 
     def setUp(self):
         pass
-
-    def test_random_point(self):
-        numpy.random.seed(0)
-        for n in range(70):
-            x, y = random_point(100, 200)
-            self.assertTrue(x >= 0,  "x is within boundaries")
-            self.assertTrue(x < 100, "x is within boundaries")
-            self.assertTrue(y >= 0,  "y is within boundaries")
-            self.assertTrue(y < 200, "y is within boundaries")
 
     def test_distance(self):
         self.assertAlmostEquals(22.360679774997898, distance((0, 0), (10, 20)))
--- a/tests/basic_map_operations_test.py
+++ b/tests/basic_map_operations_test.py
@@ -1,4 +1,4 @@
-import random
+import numpy
 import unittest
 from worldengine.basic_map_operations import distance, index_of_nearest, random_point
 
@@ -9,14 +9,13 @@ class TestBasicMapOperations(unittest.TestCase):
         pass
 
     def test_random_point(self):
-        for seed in [0, 1, 27, 29, 1939, 1982, 2015]:
-            random.seed(seed)
-            for n in range(10):
-                x, y = random_point(100, 200)
-                self.assertTrue(x >= 0,  "x is within boundaries")
-                self.assertTrue(x < 100, "x is within boundaries")
-                self.assertTrue(y >= 0,  "y is within boundaries")
-                self.assertTrue(y < 200, "y is within boundaries")
+        numpy.random.seed(0)
+        for n in range(70):
+            x, y = random_point(100, 200)
+            self.assertTrue(x >= 0,  "x is within boundaries")
+            self.assertTrue(x < 100, "x is within boundaries")
+            self.assertTrue(y >= 0,  "y is within boundaries")
+            self.assertTrue(y < 200, "y is within boundaries")
 
     def test_distance(self):
         self.assertAlmostEquals(22.360679774997898, distance((0, 0), (10, 20)))
--- a/worldengine/basic_map_operations.py
+++ b/worldengine/basic_map_operations.py
@@ -1,5 +1,4 @@
 import math
-import numpy
 
 
 def distance(pa, pb):
--- a/worldengine/basic_map_operations.py
+++ b/worldengine/basic_map_operations.py
@@ -1,9 +1,12 @@
 import math
-import random
+import numpy
 
 
-def random_point(width, height):
-    return random.randrange(0, width), random.randrange(0, height)
+def random_point(width, height, randint_substitute=None):
+    if randint_substitute is None:
+        return numpy.random.randint(0, width), numpy.random.randint(0, height)
+    else:
+        return randint_substitute(0, width), randint_substitute(0, height)
 
 
 def distance(pa, pb):
--- a/worldengine/basic_map_operations.py
+++ b/worldengine/basic_map_operations.py
@@ -2,13 +2,6 @@ import math
 import numpy
 
 
-def random_point(width, height, randint_substitute=None):
-    if randint_substitute is None:
-        return numpy.random.randint(0, width), numpy.random.randint(0, height)
-    else:
-        return randint_substitute(0, width), randint_substitute(0, height)
-
-
 def distance(pa, pb):
     ax, ay = pa
     bx, by = pb
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -391,8 +391,11 @@ def main():
 
     if args.seed is not None:
         seed = int(args.seed)
+        if not 0 <= seed <= 65535:  # platec might not be able to handle integers larger than uint16; TODO: check this
+            print("Seed has to be in the range of 0 and 65535.")
+            seed = seed % 65535
     else:
-        seed = numpy.random.randint(0, 999999)  # RNG initialization is done automatically
+        seed = numpy.random.randint(0, 65535)  # first-time RNG initialization is done automatically
     numpy.random.seed(seed)
 
     if args.world_name:
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -3,7 +3,6 @@ from argparse import ArgumentParser
 import os
 import numpy
 import pickle
-import random
 import worldengine.generation as geo
 from worldengine.common import array_to_matrix, set_verbose, print_verbose
 from worldengine.draw import draw_ancientmap_on_file, draw_biome_on_file, draw_ocean_on_file, \
@@ -393,8 +392,8 @@ def main():
     if args.seed is not None:
         seed = int(args.seed)
     else:
-        seed = random.randint(0, 65535)#RNG initialization is done automatically
-    random.seed(seed)
+        seed = numpy.random.randint(0, 999999)  # RNG initialization is done automatically
+    numpy.random.seed(seed)
 
     if args.world_name:
         world_name = args.world_name
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -389,11 +389,12 @@ def main():
         if not os.path.exists(args.FILE):
             usage("The specified world file does not exist")
 
+    maxseed = 65535  # there is a hard limit somewhere so seeds outside the uint16 range are considered unsafe
     if args.seed is not None:
         seed = int(args.seed)
-        assert 0 <= seed <= 65535, "Seed has to be in the range between 0 and 65535, borders included."  # there is a hard limit somewhere so seeds outside the uint16 range are considered unsafe
+        assert 0 <= seed <= maxseed, "Seed has to be in the range between 0 and %s, borders included." % maxseed
     else:
-        seed = numpy.random.randint(0, 65535)  # first-time RNG initialization is done automatically
+        seed = numpy.random.randint(0, maxseed)  # first-time RNG initialization is done automatically
     numpy.random.seed(seed)
 
     if args.world_name:
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -391,9 +391,7 @@ def main():
 
     if args.seed is not None:
         seed = int(args.seed)
-        if not 0 <= seed <= 65535:  # platec might not be able to handle integers larger than uint16; TODO: check this
-            print("Seed has to be in the range of 0 and 65535.")
-            seed = seed % 65535
+        assert 0 <= seed <= 65535, "Seed has to be in the range between 0 and 65535, borders included."  # there is a hard limit somewhere so seeds outside the uint16 range are considered unsafe
     else:
         seed = numpy.random.randint(0, 65535)  # first-time RNG initialization is done automatically
     numpy.random.seed(seed)
--- a/worldengine/drawing_functions.py
+++ b/worldengine/drawing_functions.py
@@ -6,7 +6,6 @@ Jython
 
 import math
 import numpy
-import random
 import sys
 import time
 from worldengine.common import get_verbose
@@ -530,7 +529,7 @@ def _dynamic_draw_a_mountain(pixels, x, y, w=3, h=3):
         max_leftborder = int(bottomness * w * 1.33)
         if not last_leftborder == None:
             max_leftborder = min(max_leftborder, last_leftborder + 1)
-        leftborder = int(bottomness * w) + random.randint(-2, 2)/2
+        leftborder = int(bottomness * w) + numpy.random.randint(-2, 2)/2
         if leftborder < min_leftborder:
             leftborder = min_leftborder
         if leftborder > max_leftborder:
@@ -557,7 +556,7 @@ def _dynamic_draw_a_mountain(pixels, x, y, w=3, h=3):
         max_modx = int(bottomness * w * 1.33)
         if not last_modx == None:
             max_modx = min(max_modx, last_modx + 1)
-        modx = int(bottomness * w) + random.randint(-2, 2)/2
+        modx = int(bottomness * w) + numpy.random.randint(-2, 2)/2
         if modx < min_modx:
             modx = min_modx
         if modx > max_modx:
@@ -603,7 +602,7 @@ def draw_ancientmap(world, target, resize_factor=1,
                     sea_color=(212, 198, 169, 255),
                     draw_biome = True, draw_rivers = True, draw_mountains = True,
                     draw_outer_land_border = False, verbose=get_verbose()):
-    random.seed(world.seed * 11)  
+    rng = numpy.random.RandomState(world.seed)  # create our own random generator
 
     if verbose:
         start_time = time.time()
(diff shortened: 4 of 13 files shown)


tests/basic_map_operations_test.py
```diff
@@ -9,14 +9,13 @@ def setUp(self):
         pass
 
     def test_random_point(self):
-        for seed in [0, 1, 27, 29, 1939, 1982, 2015]:
-            random.seed(seed)
```
tcld: This was very much unnecessary unless the RNG is expected to be of very dubious quality. I simplified it.

worldengine/basic_map_operations.py
```diff
@@ -1,9 +1,12 @@
 import math
-import random
+import numpy
 
 
-def random_point(width, height):
-    return random.randrange(0, width), random.randrange(0, height)
+def random_point(width, height, randint_substitute=None):
```
tcld: I added this after working on adding a similar feature to the noise module.
Whoever calls this now can make use of its *own* RNG instead of the system-wide one.

worldengine/cli/main.py
```diff
@@ -391,8 +391,8 @@ def main():
     if args.seed is not None:
         seed = int(args.seed)
     else:
-        seed = random.randint(0, 65535)#RNG initialization is done automatically
-    random.seed(seed)
+        seed = numpy.random.randint(0, 999999)  # RNG initialization is done automatically
```
tcld: This could be undone, if it seems problematic. I didn't see a reason to restrict worldengine to 2^16 different outputs (when no seed is passed). I am not even sure if Python supports anything less than a 32 bit integer. (The reason for not choosing a higher value is the multiplication of the seed in generation.py.)

worldengine/generation.py
```diff
@@ -188,25 +182,25 @@ def generate_world(w, step):
         return w
 
     # Precipitation with thresholds
-    PrecipitationSimulation().execute(w, seed)
+    PrecipitationSimulation().execute(w, seed * 13)
```
tcld: I wanted to centralize the seed-modifications all in one place so it is immediately visible that a change happens and everything could be easily changed later.

worldengine/simulations/permeability.py
```diff
@@ -19,20 +20,18 @@ def execute(self, world, seed):
 
     @staticmethod
     def _calculate(seed, width, height):
-        random.seed(seed * 37)
-        base = random.randint(0, 4096)
-        perm = [[0 for x in range(width)] for y in range(height)]  # TODO: replace with numpy
+        rng = numpy.random.RandomState(seed)  # create our own random generator
```
tcld: Instead of permanently modifying the global RNG (e.g. numpy.random.randint()) the sub-steps now use their own RNG whenever easily possible (and they don't modify the seeds anymore due to that now happening in generation.py).

This makes the main generation mostly independent of the order in which steps are done. In addition, should a GUI be ever able to let the user modify seeds for different steps of the generation, that seed would then actually be used instead of being multiplied by an arbitrary number.

worldengine/cli/main.py
psi29a: @ftomassetti and I talked about this. It isn't a problem in Python, but in Platec. I thought its seed was a 16-bit unsigned number. Not sure if this still the case. Honestly I don't like 99999 either... I would prefer max_int.

tcld: There are those multiplications for the different algorithms generation.py, max_int would be dangerous.
I didn't know about the platec restriction - if it were true, seeds that were manually put in should be restricted to that range, too!

psi29a: If necessary, you can always use modulus (%) 65536 which will always give you a reliable 2^16 for Platec for a given seed that happens to be bigger.

tcld: Ok, I added this in to be safe. It would be good, though, if we could add a very clear comment about this. (Or just remove the restriction if there is none.)

worldengine/basic_map_operations.py
psi29a: Neat, but we don't use the extra feature in our code.  Do you see others using this function? I'm not sure as to the utility of this and others can just as easily write their own in their code.

tcld: I could remove it, but then I would have to at least add a clarifying comment.^^ If random_point() were to be called from e.g. within temperature.py, I think it would be very important that it was *not* the global RNG that were to be queried. That would remove random numbers from the "stack" and thus influence other results depending on how often it would be called. It also would make the results for temperature.py a little non-deterministic.

Is it the parameter itself that's bugging you or rather the fact that the method now is four lines long instead of one? If it's both, I will change it back and add little note.

psi29a: I just didn't see the function being called with the extra param, so I questioned its utility. :)

If you intend on using it elsewhere in the code, then no problem.

tcld: By the way: The method currently seems to be called only at a single point (outside from the test-suite): world.py, ~line 335
Maybe we could avoid the whole discussion by just completely removing it and "inlining" the code instead. Or by removing the default value of "None" and enforcing the usage of the last parameter to make the caller fully aware of what he is doing.

worldengine/generation.py
tcld: What do you think about handing in a random value instead?

Maybe something like this:
```python
sub_seeds = numpy.random.randint(0, MAX_INT, size=100)  # create 100 random numbers; keep things deterministic in case of future expansions

PrecipitationSimulation().execute(w, sub_seeds[0])
ErosionSimulation().execute(w, sub_seeds[1])
...
```

tcld:
I might change some small things about this before uploading the new blessed ancient maps. Once that is done the tests should be fine again.

tcld:
It would be nice if somebody could look over the code; I think I touched all that is necessary, but this will need some polish, I presume.

tests/basic_map_operations_test.py
```diff
@@ -8,16 +8,6 @@ class TestBasicMapOperations(unittest.TestCase):
     def setUp(self):
         pass
 
-    def test_random_point(self):
```
tcld: The method random_point() isn't used anymore so I removed it (and this test). It was an almost trivial method anyway.

worldengine/cli/main.py
```diff
@@ -395,9 +394,12 @@ def main():
 
     if args.seed is not None:
         seed = int(args.seed)
+        if not 0 <= seed <= 65535:  # platec might not be able to handle integers larger than uint16; TODO: check this
+            print("Seed has to be in the range of 0 and 65535.")
```
ftomassetti: Perhaps we can fail at this point instead of changing the seed

worldengine/generation.py
```diff
@@ -176,31 +176,45 @@ def _around(x, y, width, height):
...
         step = Step.get_by_name(step)
-    seed = w.seed
 
     if not step.include_precipitations:
         return w
 
+    # Prepare sufficient seeds for the different steps of the generation
+    rng = numpy.random.RandomState(w.seed)  # create a fresh RNG in case the global RNG is compromised (i.e. has been queried an indefinite amount of times before generate_world() was called)
+    sub_seeds = rng.randint(0, 4294967295, size=100)  # sys.maxsize didn't quite work
+    seed_dict = {
```
ftomassetti: nice idea using a dict here

worldengine/simulations/hydrology.py
```diff
@@ -57,8 +57,8 @@ def droplet(world, pos, q, _watermap):
 
         _watermap_data = numpy.zeros((world.height, world.width), dtype=float)
         for i in range(n):
-            x, y = world.random_land()
-            if True and world.precipitation['data'][y, x] > 0:
+            x, y = world.random_land()  # will return values below zero if no land exists
```
ftomassetti: could it return None instead?

worldengine/world.py
```diff
@@ -332,11 +331,12 @@ def contains(self, pos):
...
-        x, y = random_point(self.width, self.height)
-        if self.ocean[y, x]:#TODO: this method should get a safer/quicker way of finding land!
-            return self.random_land()
-        else:
-            return x, y
+        if self.ocean.all():
+            return -1, -1  # return invalid indices if there is no land at all
+        lands = numpy.invert(self.ocean)
+        lands = numpy.transpose(lands.nonzero())  # returns a list of tuples/indices with land positions
+        y, x = lands[numpy.random.randint(0, len(lands))]  # uses global RNG
```
ftomassetti: could we return None or (None, None) here when there is no land?

ftomassetti:
It looks good to me

worldengine/cli/main.py
tcld: Ok, I'll change it.

Since you're already here: Could you elaborate on the seed-limit for platec a bit? Is there really a uint16 in use?

worldengine/generation.py
```diff
@@ -176,31 +176,45 @@ def _around(x, y, width, height):
...
     if isinstance(step, str):
         step = Step.get_by_name(step)
-    seed = w.seed
 
     if not step.include_precipitations:
         return w
 
+    # Prepare sufficient seeds for the different steps of the generation
+    rng = numpy.random.RandomState(w.seed)  # create a fresh RNG in case the global RNG is compromised (i.e. has been queried an indefinite amount of times before generate_world() was called)
+    sub_seeds = rng.randint(0, 4294967295, size=100)  # sys.maxsize didn't quite work
```
tcld: I don't like the use of this fixed number at all. Does somebody have a better idea? sys.maxint (or whatever it is called) is not available for Python 3; sys.maxsize returns a number that is not compatible with numpy (which takes a uint32).

worldengine/simulations/hydrology.py
tcld: Yes, that's a lot better. Sometimes I am stuck in the C-way of doing things.^^

worldengine/world.py
tcld: Changed it.

worldengine/cli/main.py
ftomassetti: platec_api_create here is getting a long:

https://github.com/Mindwerks/plate-tectonics/blob/master/src/platecapi.hpp#L26

I think a long could have any possible size depending on the compiler, the platform etc.

Then in the python bindings we actually store the seed in an unsigned int...

https://github.com/Mindwerks/plate-tectonics/blob/master/pybindings/platec_src/platecmodule.cpp#L19

So I would suggest we stay safe and limit the seed to 65535. Or we revise the whole thing to use platform independent types, which I am not super sure we can use in the python bindings anyway.

tcld: Thanks a lot. I will add some kind of comment since the restricted range does seem a bit odd on first glance.

tcld:
After all this I ended up here:
ancientmap_48956
ancientmap_48956

It might not be immediately visible but **everything except the mountains** is the same. Before the changes to the RNG there were a lot more (small) differences. Does anybody have an idea why the mountains differ?

ftomassetti:
the mountain generation is quite complicate. We have a mask to avoid overlapping mountains and a system to determine the size of each single mountain which is based on the elevation in the area. I am not sure if it is using the world seed.

ftomassetti:
it seems we are not using any seed for the mountains BUT we are using it for the other masks. Basically we assign each pixel to some masks, ~one for biome, and then we draw that pixel accordingly. If any mask change it could affect how mountains are drawn

tcld:
I assume you are talking about this:
```python
def _draw_a_mountain(pixels, x, y, w=3, h=3):
    # mcl = (0, 0, 0, 255)  # TODO: No longer used?
    # mcll = (128, 128, 128, 255)
    mcr = (75, 75, 75, 255)
    # left edge
    for mody in range(-h, h + 1):
        bottomness = (float(mody + h) / 2.0) / w
        leftborder = int(bottomness * w)
        darkarea = int(bottomness * w / 2)
        lightarea = int(bottomness * w / 2)
        for itx in range(darkarea, leftborder + 1):
            pixels[x - itx, y + mody] = gradient(itx, darkarea, leftborder,
                                                 (0, 0, 0), (64, 64, 64))
        for itx in range(-darkarea, lightarea + 1):
            pixels[x + itx, y + mody] = gradient(itx, -darkarea, lightarea,
                                                 (64, 64, 64), (128, 128, 128))
        for itx in range(lightarea, leftborder):
            pixels[x + itx, y + mody] = (181, 166, 127, 255)  # land_color
    # right edge
    for mody in range(-h, h + 1):
        bottomness = (float(mody + h) / 2.0) / w
        modx = int(bottomness * w)
        pixels[x + modx, y + mody] = mcr
```

Yes, this *is* complicated.^^ I worked through it a couple of days ago, but there is no directly visible use of something random (and the RNG in ```draw_ancientmap()``` is already replaced with a numpy one).

tcld:
I will take another look at it tomorrow. We are very close to the goal already.^^

tcld:
I did one more thing:
diff
So the mounts are *clearly* different; but I think the background-colors might be slightly shifted, too. (Maybe I just misunderstood how ImageMagick's compare works).
Anyway, it's a nice comparison.

tcld:
I think I found the cause for the remaining problems:
```python
def _mask(world, predicate, factor):
    _mask = [[False for x in range(factor * world.width)] for y in
             range(factor * world.height)]
    for y in range(factor * world.height):
        for x in range(factor * world.width):
            xf = int(x / factor)
            yf = int(y / factor)
            if predicate((xf, yf)):
                v = len(
                    world.tiles_around((xf, yf), radius=1,
                                       predicate=predicate))
                if v > 5:
                    _mask[y][x] = v
    return _mask
```
Apparently this map only stores integers in Python 2 (???) but floats in Python 3. It was going to be replaced by numpy anyway, I hope that will fix things.

tcld:
I could use some help with the tests since I don't think my changes make them fail.

https://github.com/Mindwerks/worldengine-data/pull/5 will take care of the ancient maps-problems.

worldengine/drawing_functions.py
```diff
@@ -142,7 +140,7 @@ def _find_mountains_mask(world, factor):
                                            radius=3,
                                            predicate=world.is_mountain))
                 if v > 32:
-                    _mask[y][x] = v / 4
+                    _mask[y, x] = v / 4.0  # force conversion to float, Python 2 will *not* do it automatically
```
tcld: Not sure which behaviour was the intended one here. The ancient map output looks different, I think the mountains are spread a little thinner for the old behaviour ```int(v/4)```.

ftomassetti:
Restarting tests after the changes to worldengine data

tcld:
I consider this passed.^^ It would be nice if this could be the next merge so I can cross this off my list and move on to finishing another branch (instead of having to come back here and rebase/fix things).

worldengine/basic_map_operations.py
```diff
@@ -1,9 +1,4 @@
 import math
-import random
-
-
-def random_point(width, height):
```
ftomassetti: I think it could be nice to have methods such as `random_point` or perhaps `random_land_point` but we could add it later if needed

worldengine/cli/main.py
```diff
@@ -392,9 +391,10 @@ def main():
 
     if args.seed is not None:
         seed = int(args.seed)
+        assert 0 <= seed <= 65535, "Seed has to be in the range between 0 and 65535, borders included."  # there is a hard limit somewhere so seeds outside the uint16 range are considered unsafe
```
ftomassetti: perhaps we could use a constant here because the number is used here and below

worldengine/drawing_functions.py
```diff
@@ -530,7 +528,7 @@ def _dynamic_draw_a_mountain(pixels, x, y, w=3, h=3):
         max_leftborder = int(bottomness * w * 1.33)
         if not last_leftborder == None:
             max_leftborder = min(max_leftborder, last_leftborder + 1)
-        leftborder = int(bottomness * w) + random.randint(-2, 2)/2
+        leftborder = int(bottomness * w) + numpy.random.randint(-2, 2)/2
```
ftomassetti: this is in an unused method so it is not very relevant but perhaps it should use some seed

worldengine/plates.py
```diff
@@ -64,7 +63,7 @@ def world_gen(name, width, height, seed, num_plates=10, ocean_level=1.0,
 
     if verbose:
         start_time = time.time()
-    add_noise_to_elevation(world, random.randint(0, 4096))
+    add_noise_to_elevation(world, numpy.random.randint(0, 4096))
```
ftomassetti: is this repeatable?

worldengine/basic_map_operations.py
tcld: It was only used once. Removing it allowed for another numpy-optimization; so for now this function would have been dead code (and since it is fairly simple anyway, rewriting it would be easy).

worldengine/cli/main.py
tcld: What exactly are you suggesting? A global variable? A local one? Something else?

worldengine/simulations/irrigation.py
```diff
@@ -5,7 +5,7 @@ class IrrigationSimulation(object):
     def is_applicable(world):
         return world.has_watermap() and (not world.has_irrigation())
 
-    def execute(self, world, seed):#seed is currently not used
+    def execute(self, world, seed):
```
ftomassetti: it is fine, execute is part of an interface, in the other implementations seed is used

ftomassetti:
Looks good to me. I would just leave a possibility to @psi29a to comment on it before merging but it is agreed that this one should be merged soon.

A lot of great work on this one, thanks a lot!

worldengine/plates.py
tcld: I believe this is the very first call to the RNG, so currently yes. Some things I left to the "global" RNG since I wasn't sure how to properly use a local one.
Should I change it?

worldengine/drawing_functions.py
tcld: Yes, it should make use of the RNG used by the ancient map operation. But since it is never called at the moment I decided to be lazy and not adapt it. I will change it now.

worldengine/cli/main.py
ftomassetti: a local one would be enough, just to avoid the repetition

worldengine/plates.py
ftomassetti: no but perhaps we could add a comment. Later we could add a test and fix this.

tcld:
I hope I took proper care of your comments.

ftomassetti:
I am satisfied, as I said I would just leave @psi29a the time to comment, if he wishes to do so. If not this is good to go for me.

ftomassetti:
Let's start merging stuff

psi29a:
No need to wait on me... I'm in the middle of moving so I'm MIA for a bit.

Issue #40, opened 2015-01-25
Explain how to read the biome map


ftomassetti:
Refer to the e-mail frm Kate

ftomassetti:
Solved via #147

PR #147, opened 2015-10-23
Added initial biomes document. 

esampson:
I'm going to write up a document to outline the various biomes. Mostly I'll probably be cutting and pasting from various other sources.

Right now the document just has a chart that shows the colors and color codes for the various biomes and groups them according to Leeman's aggregation of the Holdridge life zones. I had a document at one point that spelled out what the various zones were in better detail but I can't find it at the moment. As I do I will update the doc.

### DIFF
--- a/docs/Biomes.html
+++ b/docs/Biomes.html
@@ -1,0 +1,245 @@
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2//EN">
+
+<HTML>
+<HEAD>
+	
+	<META HTTP-EQUIV="CONTENT-TYPE" CONTENT="text/html; charset=windows-1252">
+	<TITLE></TITLE>
+	<META NAME="GENERATOR" CONTENT="OpenOffice 4.1.1  (Win32)">
+	<META NAME="AUTHOR" CONTENT="Evan Sampson">
+	<META NAME="CREATED" CONTENT="20151022;19311580">
+	<META NAME="CHANGEDBY" CONTENT="Evan Sampson">
+	<META NAME="CHANGED" CONTENT="20151023;9575539">
+	
+	<STYLE>
+		<!-- 
+		BODY,DIV,TABLE,THEAD,TBODY,TFOOT,TR,TH,TD,P { }
+		TH, TD{ padding-left: .5em; padding-right: .5em; }
+		 -->
+	</STYLE>
+	
+</HEAD>
+
+<BODY TEXT="#000000">
+<TABLE ALIGN="CENTER" CELLSPACING=0 COLS=4 RULES=NONE BORDER=0 PADDING=1em>
+	<TBODY>
+		<TR>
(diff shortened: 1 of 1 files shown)


psi29a:
It is great that you did this, but I would rather not have binary data in the repo.

Is there another way to store the data besides PDF (proprietary format that it is)?  Like LaTeX or even html?

Can the data go into the wiki?

esampson:
Sure. I only used PDF because I thought it would be more convenient. It produces a very standard output. HTML was my second choice.

psi29a:
I want to eventually use Sphinx to handle our documentation. Sphinx uses RST (reStructuredText) and very similar to markdown. With this, we can have http://readthedocs.org host our documentation and produce PDFs, LaTeX and HTML for us on every release.

For now, HTML is fine until we get sphinx setup. :)

Is it possible to split your html into two files, the markup itself and CSS? If it isn't possible, no worries. Thanks again for the work! :)

psi29a:
Crap, I still the pdf in your commit. Sorry to be a bother, but can you create a new branch but only with your last commit please?  :)

esampson:
My second commit doesn't remove the PDF? Weird.

I can split the file into HTML and CSS but I was trying to keep it a single file. Of course I may be adding some pictures later on and that will require multiple files anyway, so I suppose there's no real point to that.

If you can just confirm that the second commit isn't removing the PDF then I will just create a new branch.

coder-molok:
( Hi all, I follow this project from over a year without time to spend for
you... now, I hope this will be usefull... )

You can embed images in html file with 'data' URI.
See how here: https://en.wikipedia.org/wiki/Data_URI_scheme

My second commit doesn't remove the PDF? Weird.

I can split the file into HTML and CSS but I was trying to keep it a single
file. Of course I may be adding some pictures later on and that will
require multiple files anyway, so I suppose there's no real point to that.

If you can just confirm that the second commit isn't removing the PDF then
I will just create a new branch.

—
Reply to this email directly or view it on GitHub
<https://github.com/Mindwerks/worldengine/pull/147#issuecomment-150828784>.

psi29a:
This PR has two commits, first commit is with pdf and the second removes it and adds html. So the total PR contains both files.  If you can remove the first commit (a bit of git dark-magic) that would work, but the easiest is to just create a new branch with just the last commit (HTML) file. Then send another PR and close this.

ftomassetti:
This would close #40

esampson:
Ok. Looks like I figured out the proper git commands.

psi29a:
Awesome, merged! :)

tcld:
I am having trouble with this, too:
```
lists of files in version control and sdist do not match!

missing from sdist:

  docs

  docs/Biomes.html

suggested MANIFEST.in rules:

  recursive-include docs *.html

ERROR: InvocationError: '/home/travis/build/Mindwerks/worldengine/.tox/manifest/bin/check-manifest'

___________________________________ summary ____________________________________

ERROR:   manifest: commands failed
```

Something has to be added somewhere. Yes, that's all I know.^^

psi29a:
add this to manifest.in

```
recursive-exclude docs *.html
```

Issue #60, opened 2015-03-21
Ocean floor have a lot of noise applied to them


cesarramirez:
The ocean floor have very heavy noise applied to them. This looks very unrealistic since the ocean floor is much more flat. I see how we need the noise to make the surfaces more realistic but we may want to reduce the amount of noise applied to the ocean surface compared to the land.

I am attaching this pictures to show what I am taking about:
http://imgur.com/ShhMnV2,FOhmUey#0
http://imgur.com/ShhMnV2,FOhmUey#1

ftomassetti:
Hi Cesar, you are right. The problem should be limited using the new version of PyPlatec (about to be released). After that we will see how the situation improves and dig more into this. Thanks for reporting it.

ftomassetti:
Close via #150

ftomassetti:
Thanks for reporting this one @cesarramirez , it took a while but we fixed. Soon we will release the new version of WorldEngine with this fix and many, many improvements!

PR #150, opened 2015-10-25
Armonize ocean floor

ftomassetti:
This is a simple approach to solve the issue of noisy ocean floors (issue #60)

### DIFF
--- a/tests/blessed_images/generated_blessed_images.py
+++ b/tests/blessed_images/generated_blessed_images.py
@@ -37,6 +37,6 @@ def main(blessed_images_dir, tests_data_dir):
     img.complete()
 
 if __name__ == '__main__':
-    blessed_images_dir = os.path.dirname(os.path.realpath(__file__))
+    blessed_images_dir = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), "../../../worldengine-data/tests/images"))
     tests_data_dir = os.path.abspath(os.path.join(blessed_images_dir, '../data'))
     main(blessed_images_dir, tests_data_dir)
--- a/worldengine/generation.py
+++ b/worldengine/generation.py
@@ -141,21 +141,17 @@ def harmonize_ocean(ocean, elevation, ocean_level):
     The goal of this function is to make the ocean floor less noisy.
     The underwater erosion should cause the ocean floor to be more uniform
     """
-    height, width = elevation.shape
 
     shallow_sea = ocean_level * 0.85
     midpoint = shallow_sea / 2.0
 
-    for y in range(height):
-        for x in range(width):
-            e = elevation[y, x]
-            if ocean[y, x] and e < shallow_sea:
-                # TODO: use numpy to avoid the loop
-                if e < midpoint:
-                    e = midpoint - ((midpoint - e) / 5.0)
-                else:
-                    e = midpoint + ((e - midpoint) / 5.0)
-                elevation[y, x] = e
+    ocean_points = numpy.logical_and(elevation < shallow_sea, ocean)
+
+    shallow_ocean = numpy.logical_and(elevation < midpoint, ocean_points)
+    elevation[shallow_ocean] = midpoint - ((midpoint - elevation[shallow_ocean]) / 5.0)
+
+    deep_ocean = numpy.logical_and(elevation > midpoint, ocean_points)
+    elevation[deep_ocean] = midpoint + ((elevation[deep_ocean] - midpoint) / 5.0)
 
 # ----
 # Misc
--- a/worldengine/generation.py
+++ b/worldengine/generation.py
@@ -130,13 +130,13 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
             ('plain', hl),
             ('hill', ml),
             ('mountain', None)]
-    armonize_ocean(ocean, e, ocean_level)
+    harmonize_ocean(ocean, e, ocean_level)
     world.set_ocean(ocean)
     world.set_elevation(e, e_th)
     world.sea_depth = sea_depth(world, ocean_level)
 
 
-def armonize_ocean(ocean, elevation, ocean_level):
+def harmonize_ocean(ocean, elevation, ocean_level):
     """
     The goal of this function is to make the ocean floor less noisy.
     The underwater erosion should cause the ocean floor to be more uniform
--- a/worldengine/generation.py
+++ b/worldengine/generation.py
@@ -150,6 +150,7 @@ def harmonize_ocean(ocean, elevation, ocean_level):
         for x in range(width):
             e = elevation[y, x]
             if ocean[y, x] and e < shallow_sea:
+                # TODO: use numpy to avoid the loop
                 if e < midpoint:
                     e = midpoint - ((midpoint - e) / 5.0)
                 else:
--- a/worldengine/generation.py
+++ b/worldengine/generation.py
@@ -130,10 +130,32 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
             ('plain', hl),
             ('hill', ml),
             ('mountain', None)]
+    armonize_ocean(ocean, e, ocean_level)
     world.set_ocean(ocean)
     world.set_elevation(e, e_th)
     world.sea_depth = sea_depth(world, ocean_level)
 
+
+def armonize_ocean(ocean, elevation, ocean_level):
+    """
+    The goal of this function is to make the ocean floor less noisy.
+    The underwater erosion should cause the ocean floor to be more uniform
+    """
+    width = len(elevation[0])
+    height = len(elevation)
+
+    shallow_sea = ocean_level * 0.85
+    midpoint = shallow_sea / 2.0
+
+    for y in range(height):
+        for x in range(width):
+            e = elevation[y][x]
+            if e < shallow_sea:
+                if e < midpoint:
+                    e = midpoint - ((midpoint - e) / 5.0)
+                else:
+                    e = midpoint + ((e - midpoint) / 5.0)
+                elevation[y][x] = e
 
 # ----
 # Misc
--- a/worldengine/generation.py
+++ b/worldengine/generation.py
@@ -141,20 +141,20 @@ def armonize_ocean(ocean, elevation, ocean_level):
     The goal of this function is to make the ocean floor less noisy.
     The underwater erosion should cause the ocean floor to be more uniform
     """
-    width, height = elevation.shape
+    height, width = elevation.shape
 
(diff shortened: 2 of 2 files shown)


worldengine/generation.py
```diff
@@ -134,11 +134,33 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
...
     world.set_elevation(e, e_th)
     world.sea_depth = sea_depth(world, ocean_level)
 
 
+def armonize_ocean(ocean, elevation, ocean_level):
+    """
+    The goal of this function is to make the ocean floor less noisy.
+    The underwater erosion should cause the ocean floor to be more uniform
+    """
+    width = len(elevation[0])
```
tcld: elevation by now is a numpy-array. I would prefer this:
```python
height, width = elevation.shape
```

ftomassetti: sure, fixed

ftomassetti:
@tcld I fixed the PR according to your suggestion

worldengine/generation.py
```diff
@@ -134,11 +134,33 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
...
+    height = len(elevation)
+
+    shallow_sea = ocean_level * 0.85
+    midpoint = shallow_sea / 2.0
+
+    for y in range(height):
+        for x in range(width):
+            e = elevation[y][x]
```
tcld: Even though it is not that important, I chose to make use of ```[y, x]``` wherever possible - on one hand it is a bit faster (according to tests ~30%, which of course doesn't matter when compared to the whole generation process) and on the other hand I like its indicational effect since it is immediately visible that elevation has already been converted to numpy and thus can be used accordingly. (There probably is a numpy-way to do this loop, too. So maybe add a TODO-comment for later?)

The same goes for line 162.

worldengine/generation.py
```diff
@@ -130,11 +130,32 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
...
 
 
+def armonize_ocean(ocean, elevation, ocean_level):
+    """
+    The goal of this function is to make the ocean floor less noisy.
+    The underwater erosion should cause the ocean floor to be more uniform
+    """
+    width, height = elevation.shape
+
+    shallow_sea = ocean_level * 0.85
```
tcld: ocean_level is interpreted as "height of the ocean", right? (whereas platec uses "surface covered by ocean")
Can ocean_level be negative and throw strange results at this point? It depends on whether platec can return heightmaps with values below zero, I guess.

worldengine/generation.py
```diff
@@ -130,11 +130,32 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
...
     world.set_elevation(e, e_th)
     world.sea_depth = sea_depth(world, ocean_level)
 
 
+def armonize_ocean(ocean, elevation, ocean_level):
+    """
+    The goal of this function is to make the ocean floor less noisy.
+    The underwater erosion should cause the ocean floor to be more uniform
+    """
+    width, height = elevation.shape
```
tcld: I literally meant ```height, width = elevation.shape```. That's how all arrays are arranged.

worldengine/generation.py
tcld: There is another map called "sea_depth". Might that be useful here?

tcld:
I just noticed that the ocean-map is not used, I think the function works on the whole generated world. Is that intended?

psi29a:
Is armonize missing an h? ;)

worldengine/generation.py
rbb: Assuming elevation is a numpy array, I think the way to re-work the loop with array operations is something like:

lt = elevation< midpoint
elevation[lt] = midpoint - ((midpoint - elevation[lt) / 5.0)

Then, similar for the greater than case.

worldengine/generation.py
ftomassetti: thanks for catching this! Fixing

worldengine/generation.py
ftomassetti: this should be the elevation which corresponds to the sea level. Then we could have depressions which have an elevation below the sea level but are land. sea_depth would normally be equal to `sea_level - elevation`, of course only for actual sea/ocean tiles.

You made me realize that we are also touching tiles which are not ocean... fixing.

worldengine/generation.py
ftomassetti: yes but I need also to check if the the corresponding entry in ocean is set to true. Do you know how can I do that?

tcld: There is probably a way using numpy.ma.array (i.e. masked arrays where self.ocean itself is the mask). I did not yet manage to wrap my head around it completely; a TODO might do it for now as the complexity here is only width*height, which should not contribute that much to the total generation time.

ftomassetti:
@psi29a I was wondering if that was the case... I just saw the someone over the internet used the word "armonize" like I did but perhaps he was a creative foreigner like me :)

tcld:
If you could add a ```#TODO: numpy``` or similar, I would say this is fine. And maybe add an 'h' in front of 'armonize' unless that actually is a word I don't know. (As psi said.^^)

ftomassetti:
renamed

ftomassetti:
TODO added

tcld:
I have to admit that I don't fully understand what the loop body does. It looks to me like it isn't as much smoothing the ocean floor as it is making it shallower (below midpoint) respectively deeper (above midpoint).

worldengine/generation.py
rbb: I'll take a look tonight, after work.

ftomassetti:
@tcld it affects only ocean points which are deep enough (it ignores point above shallow_sea) and what it does it make the elevation very close to midpoint. In this way we have no huge variations in the elevation for ocean tiles. You can someone look at it as a cheap way to perform antialias.

tcld:
But it *does* make the ocean much more shallow where it was deep before, doesn't it? Maybe we should instead make use of the anti_alias()-method worldengine already heavily uses? It is a very expensive method but it is kind of made for this job.

ftomassetti:
Well we would need to adapt it to use a mask. With this method the deepest points should remain deeper than midpoint and the most shallow should remain higher than midpoint, we just reduce the variation. We could rescale it after to have a lower absolute value but I am not sure that would be helpful.

rbb:
Looks like numpy.logical_and() is the key...


    import numpy as np

    def harmonize_ocean(ocean, elevation, ocean_level):
        """
        The goal of this function is to make the ocean floor less noisy.
        The underwater erosion should cause the ocean floor to be more uniform
        """
        height, width = elevation.shape

        shallow_sea = ocean_level * 0.85
        midpoint = shallow_sea / 2.0

        for y in range(height):
            for x in range(width):
                e = elevation[y, x]
                if ocean[y, x] and e < shallow_sea:
                    # TODO: use numpy to avoid the loop
                    if e < midpoint:
                        e = midpoint - ((midpoint - e) / 5.0)
                    else:
                        e = midpoint + ((e - midpoint) / 5.0)
                    elevation[y, x] = e

    def harmonize_ocean_np(ocean, elevation, ocean_level):
        """
        The goal of this function is to make the ocean floor less noisy.
        The underwater erosion should cause the ocean floor to be more uniform
        """

        shallow_sea = ocean_level * 0.85
        midpoint = shallow_sea / 2.0

        lt = elevation_np < shallow_sea
        a = np.logical_and(lt, ocean)
        elevation[a] = midpoint - ((midpoint - elevation[a]) / 5.0)

        gt = elevation_np > shallow_sea
        a = np.logical_and(lt, ocean)
        elevation[a] = midpoint - ((midpoint - elevation[a]) / 5.0)


    ocean = np.random.rand(2,2) > 0.5;
    elevation = np.random.rand(2,2);
    elevation_np = elevation
    ocean_level = float(np.random.rand(1,1));

    print "ocean: "
    print ocean
    print "elevation: "
    print elevation
    print "ocean_level = " +str(ocean_level)

    harmonize_ocean(ocean, elevation, ocean_level);
    print "--------------------------------"
    print "elevation: "
    print elevation


    harmonize_ocean_np(ocean, elevation_np, ocean_level);
    print "--------------------------------"
    print "elevation_np: "
    print elevation_np

So, on this very small sample data, I got the same result between the two versions of code:

    In [33]: execfile('harm.py')
    ocean: 
    [[ True False]
     [False False]]
    elevation: 
    [[ 0.78725756  0.86435022]
     [ 0.69927748  0.13072007]]
    ocean_level = 0.345082907826
    --------------------------------
    elevation: 
    [[ 0.78725756  0.86435022]
     [ 0.69927748  0.13072007]]
    --------------------------------
    elevation_np: 
    [[ 0.78725756  0.86435022]
     [ 0.69927748  0.13072007]]

tcld:
Ah, this is a great idea. Although I think the heart of your code (and the parts @ftomassetti would be interested in) should be slightly altered:
```python
    valid = numpy.logical_and(elevation_np < shallow_sea, ocean)  # all tiles that need to be worked on

    lt = numpy.logical_and(elevation_np < midpoint, valid)
    elevation[lt] = midpoint - ((midpoint - elevation[lt]) / 5.0)

    gt = numpy.logical_and(elevation_np >= midpoint, valid)
    elevation[gt] = midpoint + ((elevation[gt] - midpoint) / 5.0)
```
There were some mistakes in your version. I also renamed the variables and tried to make clear what is being done.

Your small example passed since you didn't move below *midpoint*. The shallower waters was handled correctly by your code.

rbb:
I hadn't caught the elevation < shallow_sea part, but I realize the missing >= part. 

Good catches!

ftomassetti:
Thank you guys, I tried incorporating your suggestions (a.k.a. copying and pasting the code you gently provided...)

ftomassetti:
Ok, blessed images fixed, script for generating blessed images corrected. Time to merge (also because otherwise tests would fail given I have updated the blessed images)

psi29a:
Awesome work guys! :)

tests/blessed_images/generated_blessed_images.py
```diff
@@ -37,6 +37,6 @@ def main(blessed_images_dir, tests_data_dir):
     img.complete()
 
 if __name__ == '__main__':
-    blessed_images_dir = os.path.dirname(os.path.realpath(__file__))
+    blessed_images_dir = os.path.abspath(os.path.join(os.path.dirname(os.path.realpath(__file__)), "../../../worldengine-data/tests/images"))
```
tcld: I know I'm a bit late but I just noticed this:
Works for me, but seems a little too hard-coded to me. Is there maybe a better way that can be patched in?

PR #153, opened 2015-10-26
support hdf5 serialization/unserialization of all the world data

### LINKED ISSUE #91: Support HDF5
We should be able to read and write worlds in the HDF5 format

ftomassetti:
Fix #91


tcld:
I have not checked every single line, it's a bit too late for that (here). But I love it anyway.^^

What are the major advantages of this format, by the way? I know that pickle isn't fully compatible between Python 2 and 3, and I read about you pointing out some problems with Protobuf (which I honestly don't like that much either, because I believe that Google is in charge of it - and Google tends to change course by 180 degrees regardless of how many million lifes are on the line xP; I hate relying on their work).

ftomassetti:
Honestly is just that I want open worlds from Java and using protobuf I cannot do it right now. HDF5 seems a good format, with a great support for many languages and Bret suggested it for a while so I thought... why not?

ftomassetti:
tox was not aware of h5py. Let's try again

ftomassetti:
but sure, I have to install also the library... for both Travis and Appveyor. This is going to be a lot of fun, I am sure.

psi29a:
Yeah... new libraries are hell.

worldengine/hdf5_serialization.py
```diff
@@ -0,0 +1,242 @@
...
+
+    elevation_grp = f.create_group("elevation")
+    elevation_ths_grp = elevation_grp.create_group("thresholds")
+    elevation_ths_grp["sea"] = world.elevation['thresholds'][0][1]
+    elevation_ths_grp["plain"] = world.elevation['thresholds'][1][1]
+    elevation_ths_grp["hill"] = world.elevation['thresholds'][2][1]
+    elevation_data = elevation_grp.create_dataset("data", (world.height, world.width), dtype='float')
+    for y in range(world.height):
+        for x in range(world.width):
+            elevation_data[y, x] = world.elevation['data'][y][x]
```
tcld: I'm not sure how it works, but: it seems like both of these are numpy-arrays - is it really necessary to copy them element by element instead of using ```numpy.copy()``` or something similar?
If not, there are some more arrays that are copied that way (further down: ocean, sea_depth, ... - maybe even the biome-map).

worldengine/hdf5_serialization.py
```diff
@@ -0,0 +1,242 @@
...
+    general_grp["name"] = world.name
+    general_grp["width"] = world.width
+    general_grp["height"] = world.height
+
+    elevation_grp = f.create_group("elevation")
+    elevation_ths_grp = elevation_grp.create_group("thresholds")
+    elevation_ths_grp["sea"] = world.elevation['thresholds'][0][1]
+    elevation_ths_grp["plain"] = world.elevation['thresholds'][1][1]
+    elevation_ths_grp["hill"] = world.elevation['thresholds'][2][1]
+    elevation_data = elevation_grp.create_dataset("data", (world.height, world.width), dtype='float')
```
tcld: Just for the sake of consistency: I wasn't aware that one could pass a string to dtype, I always used ```dtype=float``` - and that worked fine.

tcld:
I think this looks good - although it might be possible to get rid of almost all element-by-element copy-processes that are being done, since all participants are numpy-arrays anyway. Maybe I am overlooking something there?

Other than looking at it I think that a field test is the easiest way to see if all data is properly written and read.

tcld:
What's up with the future replacement of pickle and protobuf?^^

psi29a:
I don't see this replacing protobuf, since protobuf will eventually support java.  HDF5 is just a stop-gap.

ftomassetti:
Update status: the code is done and the round-trip works (serialization -> unserialization -> back to the original world). Then I started thinking about having to install another library on AppVeyor, having to explain to users how to installed. So I cried a little, opened a bottle of whiskey and not touched this PR for a while...

Now that I am sober again I am not sure... It is true that we could make hdf5 optional as GDAL is. What do you think guys?

tests/serialization_test.py
```diff
@@ -60,6 +61,41 @@ def test_protobuf_serialize_unserialize(self):
...
+            self.assertTrue(_equal(w.precipitation,     unserialized.precipitation))
+            self.assertTrue(_equal(w.temperature,       unserialized.temperature))
+            self.assertTrue(_equal(w.sea_depth,         unserialized.sea_depth))
+            self.assertEquals(w.seed,                   unserialized.seed)
+            self.assertEquals(w.n_plates,               unserialized.n_plates)
+            self.assertTrue(_equal(w.ocean_level,       unserialized.ocean_level))
+            self.assertTrue(_equal(w.lake_map,          unserialized.lake_map))
+            self.assertTrue(_equal(w.river_map,         unserialized.river_map))
+            self.assertEquals(w.step,                   unserialized.step)
+            self.assertEqual(sorted(dir(w)),            sorted(dir(unserialized)))
```
tcld: The comparison part should be moved to its own function. I think this piece of code exists three times now. And should another save-format ever be supported, things would get even messier.

tests/serialization_test.py
```diff
@@ -60,6 +61,41 @@ def test_protobuf_serialize_unserialize(self):
...
+            self.assertEqual(w.elevation['thresholds'], unserialized.elevation['thresholds'])
+            self.assertTrue(_equal(w.ocean,             unserialized.ocean))
+            self.assertTrue(_equal(w.biome,             unserialized.biome))
+            self.assertTrue(_equal(w.humidity['quantiles'], unserialized.humidity['quantiles']))
+            self.assertTrue(_equal(w.humidity['data'],  unserialized.humidity['data']))
+            self.assertTrue(_equal(w.humidity,          unserialized.humidity))
+            self.assertTrue(_equal(w.irrigation,        unserialized.irrigation))
+            self.assertTrue(_equal(w.permeability,      unserialized.permeability))
+            self.assertTrue(_equal(w.watermap,          unserialized.watermap))
+            self.assertTrue(_equal(w.precipitation['thresholds'], unserialized.precipitation['thresholds']))
```
tcld: The _equal() method works recursively, so these manual "extractions" shouldn't be necessary (the other formats don't do it, either).

ftomassetti: yes, but when something is wrong I can find out immediately what it is. That is why I prefer to have also the single fields comparisons. For example this line:
```
self.assertTrue(_equal(w.humidity,          unserialized.humidity))
```
would make these lines redundant:
```
self.assertTrue(_equal(w.humidity['quantiles'], unserialized.humidity['quantiles']))
self.assertTrue(_equal(w.humidity['data'],  unserialized.humidity['data']))
```

However these lines tell me if I had a problem serializing/unserializing the humidity data or the humidity quantiles

tcld:
As long as it isn't trivial to add another format, I would actually prefer seeing us focus on exactly one. Pickle is kind of worthless as long as Python 2 and 3 are supported since the change in Unicode-support breaks save-compatibility between the Python versions. Protobuf has been at the latest alpha for quite a while and I personally don't like relying on Googles code - they have scrapped or "reimagined" other software before, no matter how many people relied on it. (Of course, we could stick to the latest working version in that case. And even modify it ourselves: https://github.com/google/protobuf )

So while I don't know if we need a third way of saving data, we could really use one that is actually reliable and works for all purposes we can come up with. (And then hopefully get rid of the other formats.)

tests/serialization_test.py
tcld: I think that might be a valid reason. Although this certainly makes maintenance more difficult and the gain isn't too big.

psi29a:
I think hdf5 should be optional like gdal, it is there and useful for when the right libs are in place. :)

ftomassetti:
rebased

worldengine/hdf5_serialization.py
ftomassetti: I tried this:
```
numpy.copyto(elevation_data, world.elevation['data'])
```
But I got this:
```
TypeError: argument 1 must be numpy.ndarray, not Dataset
```
I guess it is not exactly a numpy array

ftomassetti: But I can use:
`elevation_data.write_direct(world.elevation['data'])`

worldengine/hdf5_serialization.py
ftomassetti: fixed

worldengine/hdf5_serialization.py
```diff
@@ -0,0 +1,220 @@
...
+
+    elevation_grp = f.create_group("elevation")
+    elevation_ths_grp = elevation_grp.create_group("thresholds")
+    elevation_ths_grp["sea"] = world.elevation['thresholds'][0][1]
+    elevation_ths_grp["plain"] = world.elevation['thresholds'][1][1]
+    elevation_ths_grp["hill"] = world.elevation['thresholds'][2][1]
+    elevation_data = elevation_grp.create_dataset("data", (world.height, world.width), dtype=float)
+    elevation_data.write_direct(world.elevation['data'])
+
+    plates_data = f.create_dataset("plates", (world.height, world.width), dtype=int)
```
tcld: Just a question since I don't know: Is this "int" the same for Python 2 and 3?

ftomassetti: I would hope so but I do not know. Should we use something like uint16?

tcld: I don't know. Unless the tests fail, this will be fine. (Can you run the tests on Python 2 and 3?)

psi29a: I'm glad @tcld found this, yes we should avoid python datatypes when working with numpy and use numpy's standard datatypes:
https://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.html

```
numpy.dtype(numpy.uint16)
dtype('uint16')
```

ftomassetti:
In the meantime Travis complains:
```
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-Lorf5r/pygdal
ERROR: could not install deps [coverage, numpy, pygdal==1.10.0.1, Pillow, PyPlatec, noise, nose, protobuf, six, h5py]; v = InvocationError('/home/travis/build/Mindwerks/worldengine/.tox/py27/bin/pip install coverage numpy pygdal==1.10.0.1 Pillow PyPlatec noise nose protobuf six h5py (see /home/travis/build/Mindwerks/worldengine/.tox/py27/log/py27-1.log)', 1)
```

worldengine/hdf5_serialization.py
ftomassetti: fixed. I am now using numpy.uint16, numpy.float and numpy.bool

psi29a:
Numpy failed to compile... I'm going to assume that the wheel is not being installed here?

ftomassetti:
That could be the case but I do not understand why this patch is affecting numpy

psi29a:
It's not... likely a new release of numpy and we don't specify a specific version.

ftomassetti:
Should we specify a min and max release?

tcld:
The [latest version](http://sourceforge.net/projects/numpy/files/NumPy/) is about a month old.

Why would it cause problems if there was a new version of a module? Does it maybe have to be recompiled?

ftomassetti:
Perhaps it is trying to use a version for which there are no pre-compiled packages. Damn how much easier are these things on the JVM....

ftomassetti:
rebased after #187 has been merged

worldengine/hdf5_serialization.py
```diff
@@ -0,0 +1,220 @@
...
+            for x in range(world.width):
+                biome_data[y, x] = biome_name_to_index(world.biome[y][x])
+
+    if hasattr(world, 'humidity'):
+        humidity_grp = f.create_group("humidity")
+        humidity_quantiles_grp = humidity_grp.create_group("quantiles")
+        for k in world.humidity['quantiles'].keys():
+            humidity_quantiles_grp[k] = world.humidity['quantiles'][k]
+        humidity_data = humidity_grp.create_dataset("data", (world.height, world.width), dtype=numpy.float)
+        humidity_data.write_direct(world.humidity['data'])
```
tcld: Could we maybe think of a way that generalizes the saving of all the different maps? Iterate over the worlds variables. If there's a plain numpy-array, write it. If it's a dict, write it's key-/value-pairs.
It isn't absolutely necessary, but it would be great if a much more future-proof.

tests/serialization_test.py
tcld: I still think this code *has* to be moved into its own function to then replace the tests for all formats.

ftomassetti:
I cannot see wheel packages for Numpy 1.10.1 https://pypi.python.org/pypi/numpy

ftomassetti:
Appveyor is now happy but Travis keeps failing and I am out of ideas. Google did not help so far. I tried setting several numpy versions in tox.ini

ftomassetti:
```
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-yIncyD/pygdal
```

ftomassetti:
Ok, that it is fixed but tests are still failing

psi29a:
"AttributeError: 'Namespace' object has no attribute 'protobuf'"

time to rebase from master?

ftomassetti:
I wish was that but I have already rebased from master

ftomassetti:
Ok, this is solved now but still have 4 tests failing. 3 are failing because they refer to file in worldengine-data that were delete (e.g., biome_test.world which is referred by tests/biome_test.py).

tcld:
That's due to the data for https://github.com/Mindwerks/worldengine/pull/192 having already been merged while the PR itself isn't merged yet.

ftomassetti:
Ok, it seems good to go from a tests point of view. Feel free to point out corrections and improvements!

setup.py
```diff
@@ -25,7 +25,7 @@
     'entry_points': {
         'console_scripts': ['worldengine=worldengine.cli.main:main'],
     },
-    'install_requires': ['PyPlatec==1.4.0', 'pypng>=0.0.18', 'numpy>=1.9.2',
+    'install_requires': ['PyPlatec==1.4.0', 'pypng>=0.0.18', 'numpy>=1.9.2, <= 1.10.0.post2',
```
tcld: In `tox.ini` there is a different numpy-version used that *looks* like it is newer than the highest version here.

worldengine/cli/main.py
```diff
@@ -37,6 +42,8 @@ def generate_world(world_name, width, height, seed, num_plates, output_dir,
     with open(filename, "wb") as f:
         if world_format == 'protobuf':
             f.write(w.protobuf_serialize())
+        elif world_format == 'hdf5':
+            save_world_to_hdf5(w, filename)
```
tcld: The opened file `f` isn't used? That looks strange, even if it works.

worldengine/cli/main.py
```diff
@@ -228,6 +235,11 @@ def main():
              "a name is not provided, then seed_N.world, " +
              "where N=SEED",
         metavar="STR")
+    parser.add_argument('--hdf5', dest='hdf5',
+                        action="store_true",
+                        help="Save world file using HDF5 format. " +
+                             "Default = store using pickle format",
```
tcld: This has to changed to "protobuf" now.

worldengine/hdf5_serialization.py
```diff
@@ -0,0 +1,220 @@
+import h5py
+from worldengine.version import __version__
+from worldengine.biome import biome_name_to_index, biome_index_to_name
+from worldengine.world import World, Step
+import numpy
+
+
+def save_world_to_hdf5(world, filename):
+    f = h5py.File(filename, libver='latest', mode='w')
```
tcld: It's not worth the work right now, but I would have liked to see an abstract class that could be sub-classed whenever a new format is added. I am fairly certain that protobuf uses almost (?) the same functions - being able to access them as `protobuf.save_world("test.world")` and `hdf5.save_world("test.world")` would seem prettier to me.^^
Maybe that's something to remember for a future PR. :)

worldengine/hdf5_serialization.py
```diff
@@ -0,0 +1,220 @@
...
+
+    # Plates
+    w.set_plates(numpy.array(f['plates']))
+
+    # Ocean
+    w.set_ocean(numpy.array(f['ocean']))
+    w.sea_depth = numpy.array(f['sea_depth'])
+
+    # Biome
+    if 'biome' in f.keys():
```
tcld: This is something I also noticed in the protobuf save-function: What if this variable does actually not exist? Should we maybe handle the `else`-case and add in a dummy-array? Or is that overly cautious?

ftomassetti: The idea is that you could generate only certain aspects of the world, for example the elevation but not the biome. So in that case we would not create the data and therefore we would not save it.

tcld: I know. But here such a world might be loaded - and maybe one of these "optional" variables can be accessed somewhere else in the code and cause a crash?
Certainly not a huge problem, assuming the user kind of knows what he is doing.

ftomassetti: well, it should not cause a crash: the functions using optional data should first check if that data is available. For example all the simulations do that:

```
class ErosionSimulation(object):
    def is_applicable(self, world):
        return world.has_precipitations()
```

setup.py
ftomassetti: this is true. Here the version has been chosen to make AppVeyor happy. Version 1.10.1 has no wheels packages available. However version 1.10.1 was working on Travis. Let me check if other version works on Travis too.

worldengine/cli/main.py
ftomassetti: let's refactor this

worldengine/cli/main.py
ftomassetti: done, thanks

worldengine/hdf5_serialization.py
ftomassetti: I agree we have several refactorings to do, I think that we should focus on that as soon as we release this version

worldengine/hdf5_serialization.py
ftomassetti: I think we should create some intermediate concepts such as "Layer" and perhaps having subclasses like "LayerWithQuantiles" or "LayerWithThresholds". In this way we could streamline many parts of the code, including serialization.

ftomassetti:
Most comments addressed and all tests passing. Time to merge and move forward with the release?

tcld:
Do you want to release immediately after this PR? I wouldn't mind adding the [tests](https://github.com/Mindwerks/worldengine/pull/185) (although...not too important for a release, I guess) and maybe the [frozen oceans](https://github.com/Mindwerks/worldengine/pull/189) (which are done here, I just need to update the data).
Also...you really didn't want to put the serialization-stuff into its own function?^^

ftomassetti:
about the serialization stuff, like this it is much easier to debug and to implement: when I support a new serialization method I start serializating the different elements one by one and as I figure things out I add another layer. So for me it is much simpler like this, until we introduce classes for the single layers and we reorder the World class.

About other features, I guess we will be always working on cool stuff we want but a certain point we need to release stuff so that users can benefit from it. I would suggest we do a release very soon and after that we start doing releases more and more often.

tcld:
I think the frozen oceans are something that at least provides a visual benefit to users. And that's always good. ;P (Also: It is practically done, but nobody commented the latest version yet. I would have to adapt the HDF5-code a bit to make it work, though.)

The serialization-code should be put into its own function. I think it is very ugly that it appears twice in the same file (and we aren't talking about five lines of code) and even *differs slightly* due to personal preference of the person who wrote the respective lines of code.
Of course that's not a huge problem, especially since it's only tests we are talking about. I wouldn't object a merge at all, just wanted to put some more salt into the wound.^^

ftomassetti:
Well, there are other tasks to do before releasing the new version. Whatever it is merged before we tag release 0.19.0 can go in. Otherwise it could end up in 0.19.1.

This repository has ~700 commits. 200 commits have been added since we released version 0.18.0 (and we have not yet merged this PR). It is really time to share it with the world :)

tcld:
I would really like to see the satellite-map as a standard output for the next release, since it is such great work by @pangal. Including polar caps. Of course it's not a necessity but it might leave a good impression to have some more eye-candy.

ftomassetti:
I definitely agree with that: I love the satellite view!

tcld:
I suggest merging this PR and then https://github.com/Mindwerks/worldengine/pull/185 (which also fixes several bugs of the satellite-view - it's actually not just about tests). After that WorldEngine should be in a good state to release.
If there is another day we can wait, let's at least make the satellite-view a default. (Polar caps...well, I would prefer adding them now, since they also introduce a new variable that will be saved to the world-file. I don't know if that could cause any problems when done later.)

ftomassetti:
Let's start by merging this one then. At this point we should move to the next point of the schedule for releasing a new version which is getting the documentation ready. While we work on that we could merge more stuff but it is important to understand that it is not the end of the world if some feature ends up in the next release instead of this one.

PR #154, opened 2015-10-27
Update CHANGELOG.md

esampson:


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Version 0.19-UNRELEASED
 * Added export support using libgdal
 * Force rivers and grayscale_heightmap to PNG to avoid ambiguity of providing a 
 filename which can introduce input error.
+* Modified color elevation routines so that all land is raised to eliminate land below sea level and then elevation is normalized.
```
ftomassetti: I would say that depressions are represented as land instead of being represented as ocean and the ocean floor is normalized

tcld:
How about merging this, too?

ftomassetti:
Sure, I did not notice my comments were addressed. Merging.

PR #159, opened 2015-10-29
Rework of GDAL-export

tcld:
This originally went into https://github.com/Mindwerks/worldengine/pull/141 (which happened due to me working on too many things at once). I now separated it to make things easier for everybody.

I decided to make some rather big changes to the way GDAL export is handled: I think the commandline interface is now a little bit easier to use, and the changes to ```imex/__init__.py``` actually allow the user to make use of every format GDAL can handle while hopefully not making things more complicated than they were before.

I based this off of https://github.com/Mindwerks/worldengine/pull/146 even though it wasn't strictly necessary. I figured that discussing the changes happening here would take longer than merging the other PR anyway.
Here the link to the only commit that is relevant: https://github.com/Mindwerks/worldengine/commit/b8d098fc14df1f4eecb8266a731cc6edf7d7572c

### DIFF
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -356,18 +356,13 @@ def main():
     # -----------------------------------------------------
     export_options = parser.add_argument_group(
         "Export Options", "You can specify the formats you wish the generated output to be in. ")
-    export_options.add_argument("--export-type", dest="export_type",
-                                help="Export to a specific format such as: BMP or PNG",
-                                default="bmp")
-    export_options.add_argument("--export-bpp", dest="export_bpp", type=int,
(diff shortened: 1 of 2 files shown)


tcld:
The tests failed, something about recursion depth. On my system all tests were passed - can somebody help me out?

tcld:
I did a rebase so now the code is cleanly separated from other work.

ftomassetti:
Given @psi29a worked on GDAL stuff I would wait for his comments. This one is not urgent, right?

tcld:
This one is separated from everything else - unless somebody starts working on GDAL himself, this can wait.

psi29a:
I like what I see and makes a few things simpler.

Issue #158, opened 2015-10-28
pyflakes- and manifest-tests fail (tiny oversight with recent changes)


tcld:
There are some very minor things that should be fixed. I only noticed it when investigating the failed tests for one of my PRs - I should have been thoroughly looking at the tests in advance, I guess.

So, here the problems:
**A.** pyflakes-test fails (example: https://travis-ci.org/Mindwerks/worldengine/jobs/87939096 )
Cause: left-over variable from https://github.com/Mindwerks/worldengine/pull/152
Fix: trivial

**B.** manifest-test fails (example: https://travis-ci.org/Mindwerks/worldengine/jobs/87939098 )
Cause: ??? (something related to https://github.com/Mindwerks/worldengine/pull/147 )
Fix: ??? (probably trivial)

Could somebody just fix these two to get at least some of the optional tests back up? I could fix the first one myself, but it's probably easier to just fix both at once.

tcld:
Possible fixes here: https://github.com/Mindwerks/worldengine/pull/160

tcld:
Something is still going wrong: https://travis-ci.org/Mindwerks/worldengine/jobs/88091783

PR #160, opened 2015-10-29
Fixes to make pyflakes- and manifest-tests run again.

tcld:
Problem described here: https://github.com/Mindwerks/worldengine/issues/158

I used @psi29a 's suggestion regarding the addition to ```MANIFEST.in```.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,8 +5,8 @@ include *.txt
 recursive-include worldengine *.py
 recursive-include worldengine *.sh
 recursive-include worldengine *.proto
+recursive-include docs *.html
 prune appveyor
-prune docs
 prune tests
 prune tox.ini
 prune *.yml
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -10,3 +10,4 @@ prune tests
 prune tox.ini
 prune *.yml
 prune *.cfg
+recursive-exclude docs *.html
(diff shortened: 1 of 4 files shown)


tcld:
About the tests: The ancient map-stuff should pass once the RNG-changes are merged.

However, the manifest-tests still fail: https://travis-ci.org/Mindwerks/worldengine/jobs/88091783
Does somebody know what do do about it?

tcld:
@psi29a: Do you maybe have a hint about what to do to make the manifest-test pass again?

tcld:
I read about ```MANIFEST.in``` and the two optional tests can now be passed again. I excluded the ```docs```-directory (and thus ```Biomes.html```) in ```MANIFEST.in``` - is that the right way to go? If so, this could be merged to make two of the tests run again.

Maybe one of you can have a look at the (tiny) changes, @ftomassetti or @psi29a ?
I also wouldn't mind making the pyflakes-test a non-optional test, since it provides very clear instructions on what to change (and so far it has proven quite useful in my opinion).

ftomassetti:
I think we should include the docs directory in MANIFEST.in, so that the user downloading the package have also the docs.

ftomassetti:
Aside from including the docs all the rest seems good to me

tcld:
So I should use ```recursive-include``` instead? I am not sure if the the test will pass then, but I can try.

tcld:
Ok, it seems to work.

ftomassetti:
cool, merging

PR #161, opened 2015-10-29
numpy-replacements for image-drawing

tcld:
Another portion of the code in https://github.com/Mindwerks/worldengine/pull/141 . Some very small changes and quite a lot of changes in the draw- and drawing_functions-modules.

This again is based off of https://github.com/Mindwerks/worldengine/pull/146 .

Tests will fail for now due to some differences in rounding for the grayscale-map.

### DIFF
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -3,7 +3,7 @@ import numpy
 import numpy.ma as ma
 
 from worldengine.drawing_functions import draw_ancientmap, \
-    draw_rivers_on_image, gradient
+    draw_rivers_on_image
 
 # -------------
 # Helper values
@@ -168,43 +168,33 @@ def draw_simple_elevation(world, sea_level, target):
     """ This function can be used on a generic canvas (either an image to save
         on disk or a canvas part of a GUI)
     """
-    min_elev_sea = None
-    max_elev_sea = None
-    min_elev_land = None
-    max_elev_land = None
-    for y in range(world.height):
-        for x in range(world.width):
-            e = world.elevation['data'][y,x]
-            if sea_level is None:
-                if min_elev_land is None or e < min_elev_land:
-                    min_elev_land = e
-                if max_elev_land is None or e > max_elev_land:
-                    max_elev_land = e
-            elif world.is_land((x, y)):
-                if min_elev_land is None or e < min_elev_land:
-                    min_elev_land = e
-                if max_elev_land is None or e > max_elev_land:
-                    max_elev_land = e
-            else:
-                if min_elev_sea is None or e < min_elev_sea:
-                    min_elev_sea = e
-                if max_elev_sea is None or e > max_elev_sea:
-                    max_elev_sea = e
+    e = world.elevation['data']
 
-    elev_delta_land = (max_elev_land - min_elev_land)/11
-    if sea_level != None:
+    c = numpy.empty(e.shape, dtype=numpy.float)
+    if sea_level is None or world.ocean is None or not world.ocean.any():  # or 'not any ocean'
+        mask = numpy.ma.array(e, mask = False)  # the whole map
+        min_elev_land = mask.min()
+        max_elev_land = mask.max()
+        elev_delta_land = (max_elev_land - min_elev_land) / 11.0
+
+        c = ((e - min_elev_land) / elev_delta_land) + 1
+    else:
+        mask = numpy.ma.array(e, mask = world.ocean)  # only land
+        min_elev_land = mask.min()
+        max_elev_land = mask.max()
+        elev_delta_land = (max_elev_land - min_elev_land) / 11.0
+
+        mask = numpy.ma.array(e, mask = numpy.logical_not(world.ocean))  # only ocean
+        min_elev_sea = mask.min()
+        max_elev_sea = mask.max()
         elev_delta_sea = max_elev_sea - min_elev_sea
-    
+
+        c[world.ocean] = ((e[world.ocean] - min_elev_sea) / elev_delta_sea)
+        c[numpy.invert(world.ocean)] = ((e[numpy.invert(world.ocean)] - min_elev_land) / elev_delta_land) + 1
+
     for y in range(world.height):
         for x in range(world.width):
-            e = world.elevation['data'][y, x]
-            if sea_level is None:
-                c = ((e - min_elev_land) / elev_delta_land) + 1
-            elif world.is_land((x, y)):
-                c = ((e - min_elev_land) / elev_delta_land) + 1
-            else:
-                c = ((e - min_elev_sea) / elev_delta_sea)
-            r, g, b = elevation_color(c, sea_level)
+            r, g, b = elevation_color(c[y, x], sea_level)
             target.set_pixel(x, y, (int(r * 255), int(g * 255),
                                     int(b * 255), 255))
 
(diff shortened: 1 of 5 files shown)


worldengine/draw.py
```diff
@@ -212,46 +202,34 @@ def draw_riversmap(world, target):
...
+
+    mask = numpy.ma.array(e, mask = numpy.logical_not(world.ocean))  # only ocean
+    min_elev_sea = mask.min()
+    max_elev_sea = mask.max()
     elev_delta_sea = max_elev_sea - min_elev_sea
 
+    c = numpy.empty(e.shape, dtype=numpy.float)
+    c[numpy.invert(world.ocean)] = (e[numpy.invert(world.ocean)] - min_elev_land) * 127 / elev_delta_land + 128
+    c[world.ocean] = (e[world.ocean] - min_elev_sea) * 127 / elev_delta_sea
+    c = numpy.rint(c).astype(dtype=numpy.int32)  # proper rounding
```
tcld: This was not rounded before, which might even be preferable.

worldengine/draw.py
```diff
@@ -306,24 +284,16 @@ def draw_precipitation(world, target, black_and_white=False):
...
-                if low is None or p < low:
-                    low = p
-                if high is None or p > high:
-                    high = p
+        low = world.precipitation['data'].min()
+        high = world.precipitation['data'].max()
+        floor = 0
+        ceiling = 255
+
+        colors = numpy.interp(world.precipitation['data'], [low, high], [floor, ceiling])
```
tcld: I am making use of numpys interp-method (which basically does what drawing_functions.py/gradient() did before). I assume it is faster.
I also prepared things for the 16 Bit grayscale-output that can also be used for temperature and precipitation, if wanted.

worldengine/drawing_functions.py
```diff
@@ -590,7 +588,8 @@ def _draw_a_mountain(pixels, x, y, w=3, h=3):
         pixels[x + modx, y + mody] = mcr        
 
 
-def _pseudo_random_land_pos(world, i):
+def _pseudo_random_land_pos(world, i):  # What is this good for? Why is it not truly random?
```
tcld: What is this method actually doing that cannot be done by world.random_land(), a truly random function?

ftomassetti: this is fully repeatable

ftomassetti:
we still want to merge this, right? Rerunning the tests

tcld:
Sure.^^ I closed the PR that doesn't have to be merged, these new ones are the replacements.
I did a rebase, too, so now the changed code actually reflects what has been changed.

worldengine/drawing_functions.py
tcld: I should remove my comment then.^^ Due to the recent changes to the RNG this function could technically be removed, I guess. It would require a couple of small changes on top of it, though.

worldengine/world.py
```diff
@@ -351,7 +350,7 @@ def sea_level(self):
     # Tiles around
     #
 
-    def on_tiles_around_factor(self, factor, pos, radius=1, action=None):
+    def on_tiles_around_factor(self, factor, pos, action, radius=1):
```
tcld: ```action=None``` seemed unsafe since that case wasn't even handled by the function. In other words: ```action``` was a required parameter disguised as an optional one.

worldengine/drawing_functions.py
```diff
@@ -705,16 +691,9 @@ def on_border(pos):
...
-    max_elev = None
-    for y in range(world.height):
-        for x in range(world.width):
-            e = world.elevation['data'][y, x]
-            if min_elev is None or e < min_elev:
-                min_elev = e
-            if max_elev is None or e > max_elev:
-                max_elev = e
-    # elev_delta = max_elev - min_elev  # TODO: no longer used?
+    #min_elev = world.elevation['data'].min()
```
ftomassetti: mmm why do we keep code commented out?

worldengine/drawing_functions.py
```diff
@@ -637,64 +638,49 @@ def draw_ancientmap(world, target, resize_factor=1,
...
 
     def unset_jungle_mask(pos):
-        x, y = pos
-        jungle_mask[y][x] = False
+        jungle_mask[pos[1], pos[0]] = 0
 
     def unset_tundra_mask(pos):
-        x, y = pos
-        tundra_mask[y][x] = False
+        tundra_mask[pos[1], pos[0]] = 0
```
ftomassetti: I found the `x,y = pos` clearer

worldengine/world.py
ftomassetti: ok, have you checked all the invokation of this method, given we are changing the order of the parameters? Same goes for the following function

worldengine/drawing_functions.py
tcld: Ok, I can change that back.

worldengine/world.py
tcld: I have, and they were all called using the keyword.
Although I didn't find a call to ```on_tiles_around```.

worldengine/drawing_functions.py
tcld: It was there before, I didn't know if I could remove it. (Before the loop was still run, even though the variables weren't needed anywhere. So it looked less dead than it was.)

worldengine/drawing_functions.py
```diff
@@ -705,33 +709,15 @@ def on_border(pos):
...
-                max_elev = e
-    # elev_delta = max_elev - min_elev  # TODO: no longer used?
-    if verbose:
-        elapsed_time = time.time() - start_time
-        print(
-            "...drawing_functions.draw_oldmap_on_pixel: max, min elevation " +
-            "Elapsed time " + str(elapsed_time) + "  seconds.")
-
-    if verbose:
-        start_time = time.time()
```
tcld: This removed code was dead, the variables were not used anywhere.

tcld:
Updated.

tcld:
Except for the difference in rounding of the heightmaps (line 229 of this PR) this passes the tests and is otherwise finished. I added another PR to worldengine-data https://github.com/Mindwerks/worldengine-data/pull/8 (even though the heightmaps will be changed again when moving them to 16 Bit) and then this is as ready is it gets (aside from code-critique).

tcld:
Rebased and conflicts solved.

worldengine/draw.py
```diff
@@ -168,43 +168,33 @@ def draw_simple_elevation(world, sea_level, target):
...
-                if max_elev_sea is None or e > max_elev_sea:
-                    max_elev_sea = e
+    e = world.elevation['data']
+
+    c = numpy.empty(e.shape, dtype=numpy.float)
+    if sea_level is None or world.ocean is None or not world.ocean.any():  # or 'not any ocean'
+        mask = numpy.ma.array(e, mask = False)  # the whole map
+        min_elev_land = mask.min()
+        max_elev_land = mask.max()
+        elev_delta_land = (max_elev_land - min_elev_land) / 11.0
```
ftomassetti: nice :)

worldengine/draw.py
```diff
@@ -168,43 +168,33 @@ def draw_simple_elevation(world, sea_level, target):
...
-                if min_elev_sea is None or e < min_elev_sea:
-                    min_elev_sea = e
-                if max_elev_sea is None or e > max_elev_sea:
-                    max_elev_sea = e
+    e = world.elevation['data']
+
+    c = numpy.empty(e.shape, dtype=numpy.float)
+    if sea_level is None or world.ocean is None or not world.ocean.any():  # or 'not any ocean'
+        mask = numpy.ma.array(e, mask = False)  # the whole map
+        min_elev_land = mask.min()
```
ftomassetti: it seems much of this if code and the else code could be extracted from the if

worldengine/draw.py
```diff
@@ -168,43 +168,33 @@ def draw_simple_elevation(world, sea_level, target):
...
-    elev_delta_land = (max_elev_land - min_elev_land)/11
-    if sea_level != None:
+        c = ((e - min_elev_land) / elev_delta_land) + 1
+    else:
+        mask = numpy.ma.array(e, mask = world.ocean)  # only land
+        min_elev_land = mask.min()
+        max_elev_land = mask.max()
+        elev_delta_land = (max_elev_land - min_elev_land) / 11.0
+
+        mask = numpy.ma.array(e, mask = numpy.logical_not(world.ocean))  # only ocean
```
ftomassetti: perhaps it could be clearer to use mask_land and mask_ocean instead of reusing mask

worldengine/draw.py
```diff
@@ -213,46 +203,34 @@ def draw_riversmap(world, target):
...
     land_color = (0, 0, 0, 255)
 
-    for y in range(world.height):#TODO: numpy
+    for y in range(world.height):
         for x in range(world.width):
-            if world.ocean[y, x]:
-                target.set_pixel(x, y, sea_color)
-            else:
-                target.set_pixel(x, y, land_color)
+            target.set_pixel(x, y, sea_color if world.ocean[y, x] else land_color)
```
ftomassetti: could we use the method `is_ocean` here instead of accessing `world.ocean[y, x]`, that is because we could change the way we store ocean data later

worldengine/draw.py
```diff
@@ -307,24 +285,16 @@ def draw_precipitation(world, target, black_and_white=False):
...
-        for y in range(height):
-            for x in range(width):
-                p = world.precipitations_at((x, y))
-                if low is None or p < low:
-                    low = p
-                if high is None or p > high:
-                    high = p
+        low = world.precipitation['data'].min()
+        high = world.precipitation['data'].max()
+        floor = 0
```
ftomassetti: +1 for the variables with clear names

worldengine/drawing_functions.py
```diff
@@ -56,48 +56,48 @@ def draw_rivers_on_image(world, target, factor=1):
...
-    for y in range(world.height * factor):#TODO: numpy
+    for y in range(world.height * factor):  # TODO: numpy
         for x in range(world.width * factor):
             if world.ocean[int(y / factor), int(x / factor)]:
                 _ocean[y, x] = True
 
     def my_is_ocean(pos):
-        return _ocean[pos[1], pos[0]]
+        x, y = pos
+        return _ocean[y, x]
```
ftomassetti: nice

worldengine/simulations/basic.py
```diff
@@ -3,13 +3,19 @@
...
-        mask = map_data
-        if ocean is not None:
-            mask = numpy.ma.array(map_data, mask = ocean)
-        mask = numpy.ma.masked_less_equal(mask, e)
+    #maybe map was already masked when we got it; if not, this will make sure we operate on a mask
+    mask = numpy.ma.array(map_data, mask = False, keep_mask = True)
+
+    if ocean is not None:
+        if ocean.shape != map_data.shape:
+            raise Exception(
```
ftomassetti: Nice error message

ftomassetti:
Some minor comments but it seems good to go, sorry if it took a while to review it

worldengine/draw.py
tcld: Ok, I tried extracting the code while keeping it easy to read. ~Ten minutes until the commit is ready.

worldengine/draw.py
tcld: Ok. Uses a little more memory, I assume, but it might be easier to read.

tcld:
Changes are up, hopefully it's fine now.

tcld:
Oh, and don't forget: For this to pass the tests, worldengine-data needs to be updated (https://github.com/Mindwerks/worldengine-data/pull/8).

ftomassetti:
Ok, I ran the code locally with the patch for worldengine-data and this one and tests pass. Merging both.

Issue #140, opened 2015-10-18
Some small Ideas to discuss


tcld:
I have been thinking about some things I noticed (this is not an "issue" but I don't know of any other place for this):
Elevation maps generated by platec usually contain values between 0 and 10; although some maps go up to even 14.
* *Would it be appropriate/feasible/wanted to scale that range up?*

Advantages I see:
Imagine the elevation map being multiplied by 2000: The range between highest and lowest points of a world would be at roughly 20000 - if you translate that to meters it is pretty much as on earth, with the occasional world straying a bit further from that.
Thinking back to the strange rivers I was asking for some sense of scale to decide how much a "drop" of water can achieve - using our real physical units as a metric would be a great start for that. Instead of having to combine two quite alien scales, we would have a frame of reference to help make certain decisions.

There is a function that sets elevation thresholds based on a fixed percentage of land that should be declared e.g. a 'mounain'.
* *In conjunction with 1., would it be desirable to remove the fixed percentage and instead use a fixed "height"?*

Advantages:
Not every world has exactly three percent mountains, but instead it depends on the seed if there are mountains and hills.

I thought it would be great if worldengine could generate higher-detail heightmaps (a pull request will show up at some point). I saw (in the README maybe?) that some other projects respectively games made use of worldengines maps already. I myself initially found worldengine looking for a program to create nice heightmaps to be used in a toy-project I was working on in Unreal Engine 4.
So I dug into draw.py and the GDAL-export() and noticed that the pillow library seems to be exclusively used to write out the finished PNGs of a world - nowhere else. And I found no way to export 16 Bit grayscale images with it. (GDAL can do it, but it is a pain to install!)
* *Should pillow maybe replaced by a different image-writing library?*

Advantages:
After taking a look at several libraries I found some that can handle many if not all PNG-standards and thus seem more useful than Pillow (at least when it comes to writing PNGs) - the one I tried out is this: PyPNG (https://pythonhosted.org/pypng/ - seemingly available for Python 2.3 and above)
With just a small bit of work I made worldengine export proper 16 Bit heightmaps and even took a look at one in Unreal Engine 4. :) Seeing the world go 3D is nice - maybe it will prove useful to others!

I think that all of these are technically easy to realize, 1. would probably be the most complicated (depending on how much code depends on the scale of the elevation map). The benefits are there, but since those still are significant changes, I wanted to ask what you guys think about them. (The PNG-stuff will be visible in my next pull request to get an idea of it, but it could be removed).

[...]

PR #162, opened 2015-10-30
Temperature and Humidity parameters can be altered from the command line.

esampson:
Added command line options for Temperature and Humidity ranges. While this does not fully allow the creation of non-terrestrial planets (we need to be able to alter how much water there is) it brings us much further along.

$ python worldengine -b -s 1 --temps 1/1/1/1/0.594/0.439  --humidity .507/.236/.073/.014/.002/.0/.0
Worldengine - a world generator (v. 0.18.0)
-----------------------
 operation         : world generation
 seed                 : 1
 name                 : seed_1
 width                : 512
 height               : 512
 number of plates     : 10
 world format         : protobuf
 black and white maps : False
 step                 : full
 greyscale heightmap  : False
 rivers map           : False
 fade borders         : True
 temperature ranges   : [1.0, 1.0, 1.0, 1.0, 0.594, 0.439]
 humidity ranges      : [0.507, 0.236, 0.073, 0.014, 0.002, 0.0, 0.0]

starting (it could take a few minutes) ...

Producing ouput:
* world data saved in './seed_1.world'
* ocean image generated in './seed_1_ocean.png'
* precipitation image generated in './seed_1_precipitation.png'
* temperature image generated in './seed_1_temperature.png'
* biome image generated in './seed_1_biome.png'
* elevation image generated in './seed_1_elevation.png'
...done

ancient_map_seed_1

### DIFF
--- a/tests/serialization_test.py
+++ b/tests/serialization_test.py
@@ -11,7 +11,7 @@ class TestSerialization(unittest.TestCase):
         self.maxDiff = None
 
     def test_pickle_serialize_unserialize(self):
-        w = world_gen("Dummy", 32, 16, 1, step=Step.get_by_name("full"))
+        w = world_gen("Dummy", 32, 16, 1, [.874, .765, .594, .439, .366, .124], [.941, .778, .507, .236, 0.073, .014, .002], step=Step.get_by_name("full"))
         f = tempfile.NamedTemporaryFile(delete=False).name
         w.to_pickle_file(f)
         unserialized = World.from_pickle_file(f)
@@ -37,7 +37,7 @@ class TestSerialization(unittest.TestCase):
         self.assertEqual(w, unserialized)
 
     def test_protobuf_serialize_unserialize(self):
-        w = world_gen("Dummy", 32, 16, 1, step=Step.get_by_name("full"))
+        w = world_gen("Dummy", 32, 16, 1, [.874, .765, .594, .439, .366, .124], [.941, .778, .507, .236, 0.073, .014, .002], step=Step.get_by_name("full"))
         serialized = w.protobuf_serialize()
         unserialized = World.protobuf_unserialize(serialized)
         self.assertTrue(_equal(w.elevation['data'], unserialized.elevation['data']))
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -302,12 +302,12 @@ def main():
     g_generate.add_argument('--temps', dest='temps', 
                         help="Provide alternate ranges for temperatures. " +
                              "If not provided, the default values will be used. \n" +
-                             "[default = 0.874/0.765/0.594/0.439/0.366/0.124]",
+                             "[default = .126/.235/.406/.561/.634/.876]",
                             metavar="#/#/#/#/#/#")
     g_generate.add_argument('--humidity', dest='humids', 
                         help="Provide alternate ranges for humidities. " +
                              "If not provided, the default values will be used. \n" +
-                            "[default = .941/.778/.507/.236/.073/.014/.002]",
+                            "[default = .059/.222/.493/.764/.927/.986/.998]",
                             metavar="#/#/#/#/#/#/#")
     g_generate.add_argument('--not-fade-borders', dest='fade_borders', action="store_false",
                                help="Not fade borders",
@@ -439,7 +439,7 @@ def main():
     if args.temps:
         temps = args.temps.split('/')
         for x in range(0,6):
-            temps[x] = float(temps[x])
+            temps[x] = 1 - float(temps[x])
 
     if args.humids and not generation_operation:
         usage(error="humidity can be assigned only during world generation")
@@ -451,7 +451,7 @@ def main():
     if args.humids:
         humids = args.humids.split('/')
         for x in range(0,7):
-            humids[x] = float(humids[x])
+            humids[x] = 1 - float(humids[x])
 
     print('Worldengine - a world generator (v. %s)' % VERSION)
     print('-----------------------')
@@ -469,9 +469,9 @@ def main():
         print(' rivers map           : %s' % args.rivers_map)
         print(' fade borders         : %s' % args.fade_borders)
     if args.temps:
-        print(' temperature ranges   : %s' % temps)
+        print(' temperature ranges   : %s' % args.temps)
     if args.humids:
-        print(' humidity ranges      : %s' % humids)
+        print(' humidity ranges      : %s' % args.humids)
     if operation == 'ancient_map':
         print(' resize factor          : %i' % args.resize_factor)
         print(' world file             : %s' % args.world_file)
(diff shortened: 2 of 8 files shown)


tcld:
Interesting! Although I wouldn't know what kind of variables I would have to pass to worldengine, the explanation could be more precise. From the code I gather that these are the percentage-values that have been fixed before - your example
```--temps 1/1/1/1/0.594/0.439```
thus would result in *cool* temperature region everywhere (the ones before are overwritten), then a bit of *warm* (15.5%, the rest is overwritten) and 43.9% *subtropical*. The percentages refer to the temperature map, but I *think* this still results in e.g. 43.9% of the (lands) surface being covered by a subtropical temperature-region.

Is that right? If so, an ice-world would be ```--temps 1/0/0/0/0/0``` ?

**Comment by tcld** [comment]
On a sidenote: If your PRs aren't passing the ancient map-tests even though you think they should, rebase your work on the current master (my last PR changed the ancient map-generation slightly).

**Comment by tcld** [comment]
I just realized that this is probably not repeatable - maybe worldengines *info-mode* needs to be updated, too? (This would then require an update to the save-files, too, I guess.)

**Comment by ftomassetti** [comment]
I am not sure I get your last comment. Can you rephrase that please?

**Comment by tcld** [comment]
Sorry. Changing the parameters for temperature- and humidity-distributions will result in a world-file whose generation-parameters can not be fully retrieved for later reproduction. Maybe I am just overlooking something, or maybe the info-mode isn't meant to provide that kind of reproducibility anyway.

**Comment by ftomassetti** [comment]
Oh, ok I understand now. So we could just store those parameters in the world file. Currently I do not think it offers 100% reproducibility but it would make sense to store it.

**Comment by tcld** [comment]
We might just not store them - there may be more changes like this in the future and it might prove laborious to store all the different parameters in the world-file. Depends on how important it is to be able to recreate a world.

**Comment by esampson** [comment]
Ok. Merged with the new master and updated.

The update is very minor, but I think it will make it much easier to understand. The values I was initially using were the identical values used in code; good for the machine but not so good for human comprehension. What I did was invert the value (1 - old value) and that should (hopefully) make it a lot easier to understand what is going on.

The way the parameters work is fairly simple. Temperature is arranged in the order of coldest to hottest and Humidity is arranged driest to wettest. With the modification the following would have been to correct command:

--temps 0/0/0/0/.406/.561 --humidity .492/.764/.927/.986/.998/1/1

These numbers are the dividing lines based on the percent of the land population. In other words 49.2% of the land surface on this map will be the driest possible category. 76.4% will be either the driest or the second driest (and since 49.2% is already designated as the driest that means that 27.2% will be the next driest), etc. There are 8 possible humidity zones normally but on this map the 2 wettest won't occur because the lines separating zones 6 from 7 and 7 from 8 occur at the very end of the chart, so everything will be zone 6 or drier.

Likewise there are 7 land zone, but the 4 coldest won't occur in this case because the the lines separating zones 4 and 5 (and everything before that) occur at 0% of the population.

**Comment by esampson** [comment]
As for slightly randomizing the values; I think it would be a better idea to just add a command line option that will randomly generate new values (something like --humidity random). That way the default will be to use the tuned values but you can easily use random values if you like.

File: tests/serialization_test.py [hunk 1]
```diff
@@ -11,7 +11,7 @@ def setUp(self):
         self.maxDiff = None
 
     def test_pickle_serialize_unserialize(self):
-        w = world_gen("Dummy", 32, 16, 1, step=Step.get_by_name("full"))
+        w = world_gen("Dummy", 32, 16, 1, [.874, .765, .594, .439, .366, .124], [.941, .778, .507, .236, 0.073, .014, .002], step=Step.get_by_name("full"))
```
tcld: Might it be prettier to give world_gen default values for the new parameters so you don't have to explicitly pass them here? Should these values change (or a 7th be added) during later development, it would be a bit ugly to have to go around and fix the tests (instead of e.g. a standard parameter within world_gen).

ftomassetti:
yes, I agree with having "earth like" values by default and random as an option

tcld:
I am currently thinking about randomization and storing variables in a global file (https://github.com/Mindwerks/worldengine/issues/140#issuecomment-152483413). Something like this might be nice:
```python
temps=[.874, .765, .594, .439, .366, .124]  # all values below 1: relative (% of surface covered)
temps=[0, 2.1, 5.0, 5.7, 5.9, 10.0]  # at least 1 value above one: absolute temp-value required
temps=[.874, .765, (.594, .05), .439, .366, .124]  # a 2-tuple means: 0.594 +/- 0.050
temps=[.874, .765, (.594, +0.05, -0.1), .439, .366, .124]  # a 3-tuple means: 0.494...0.644
```
A class would have to manage these lists. My ideas regarding this are still very rough.^^

worldengine/cli/main.py
```diff
@@ -418,6 +429,30 @@ def main():
...
+    if args.humids and not generation_operation:
+        usage(error="humidity can be assigned only during world generation")
+
+    if args.humids and len(args.humids.split('/')) is not 7:
+        usage(error="humidity must have exactly 7 values")
+
+    humids = [.941, .778, .507, .236, 0.073, .014, .002]
+    if args.humids:
+        humids = args.humids.split('/')
+        for x in range(0,7):
```
tcld:
```python
for x in range(0, len(humids)):
```
Try not to hardcode too much stuff, it will be a pain to maintain.

tcld:
Thinking about the new notation you chose, to me it is as convoluted as the previous one. *Exactly* as convoluted. I would have preferred not having that extra layer of confusion (if a user looks into the code he has to notice that he has to look at values in the code differently than values on the commandline).

worldengine/cli/main.py
tcld: It seems I was too late. This still needs a fix, though. (Line 441 does, too.)

ftomassetti:
I think this is a great contribution and we can incrementally improve it. It is very good and positive to have more eyes looking at the code but perhaps we should not be too negative :)

tcld:
If one of those lists ever changes their length, 4 different files will have to be changed to account for it - two files belonging to the test-suite. Almost all of that could have been avoided by slightly altering ~5 lines of code.

I am sorry if I come across as overly negative, this is a great addition for sure. I just think that increasing maintainability is worth going the extra mile.

ftomassetti:
I appreciate your comments and I agree with many of your concerns. I just had experiences with other projects were we ended up parking PRs for a long time, then they ended up out of sync with master so we had to rebase them or new committers were scared of. This is absolute not that case, I wanted just to say that perhaps we can each time point out the main 1-2 things, merge and then refactor. In this way we reduce conflicts and keep the unit of work small, which I think is positive. But your attention to details and code quality is very much appreciated and we definitely need some help on that side!

esampson:
I was trying earlier to use default values when possible, but for some reason they didn't seem to be getting passed correctly. I'll be making more passes over this code to try and tighten and clean up bits of it (such as this) and I appreciate all of the feed back.

PR #163, opened 2015-10-31
16 Bit grayscale heightmaps

tcld:
This is the setup I am using to generate 16 Bit grayscale heightmaps, extracted from the (now dead) #141 . There is a big chunk of code here that is basically a drop-in replacement of the ImageWriter-class used so far (I gave it its own file). Technically that class could be used to write all images (which would remove the Pillow-dependency) but I didn't want to go in too deep again. Maybe later.

The tests won't be passed until I also update the grayscale heightmap in worldengine-data. That can happen once the code here is acceptable, I think.

EDIT: I had to base this off of #161 , currently only the last two commits here are actually new.

PS: I will try hard to keep this my last PR until all others are handled. ;)

### DIFF
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -819,8 +819,7 @@ def draw_ancientmap_on_file(world, filename, resize_factor=1,
                             sea_color=(212, 198, 169, 255),
                             draw_biome=True, draw_rivers=True, draw_mountains=True, 
                             draw_outer_land_border=False, verbose=False):
-    img = ImagePixelSetter(world.width * resize_factor,
-                           world.height * resize_factor, filename)
+    img = PNGWriter.rgba_from_dimensions(world.width * resize_factor, world.height * resize_factor, filename)
     draw_ancientmap(world, img, resize_factor, sea_color,
                     draw_biome, draw_rivers, draw_mountains, draw_outer_land_border, 
                     verbose)
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -801,6 +801,6 @@ def draw_scatter_plot_on_file(world, filename):
 
 
 def draw_satellite_on_file(world, filename):
-    img = PNGWriter(world.width, world.height, filename)
+    img = PNGWriter.rgba_from_dimensions(world.width, world.height, filename)
     draw_satellite(world, img)
     img.complete()
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -1,4 +1,3 @@
-from PIL import Image
 import numpy
 import numpy.ma as ma
 
@@ -315,37 +314,6 @@ def get_biome_color_based_on_elevation(world, elev, x, y):
 # Draw on generic target
 # ----------------------
 
-
-class ImagePixelSetter(object):
-
-    def __init__(self, width, height, filename):
-        self.img = Image.new('RGBA', (width, height))
-        self.pixels = self.img.load()
-        self.filename = filename
-
-    def set_pixel(self, x, y, color):
-        if len(color) == 3:  # Convert RGB to RGBA - TODO: go through code to fix this
-            color = (color[0], color[1], color[2], 255)
-        self.pixels[x, y] = color
-
-    def complete(self):
-        try:
-            self.img.save(self.filename)
-        except KeyError:
-            print("Cannot save to file `{}`, unsupported file format.".format(self.filename))
-            filename = self.filename+".png"
-            print("Defaulting to PNG: `{}`".format(filename))
-            self.img.save(filename)
-
-    def __getitem__(self, item):
-        return self.pixels[item]
-
-    def __setitem__(self, item, value):
-        if len(value) == 3:  # Convert RGB to RGBA - TODO: go through code to fix this
-            value = (value[0], value[1], value[2], 255)
-        self.pixels[item] = value
-
-
 def draw_simple_elevation(world, sea_level, target):
     """ This function can be used on a generic canvas (either an image to save
         on disk or a canvas part of a GUI)
@@ -553,7 +521,7 @@ def draw_precipitation(world, target, black_and_white=False):
         low = world.precipitation['data'].min()
         high = world.precipitation['data'].max()
         floor = 0
-        ceiling = 255
+        ceiling = 255  # could be changed into 16 Bit grayscale easily
 
         colors = numpy.interp(world.precipitation['data'], [low, high], [floor, ceiling])
         colors = numpy.rint(colors).astype(dtype=numpy.int32)  # proper rounding
@@ -603,7 +571,7 @@ def draw_temperature_levels(world, target, black_and_white=False):
         low = world.temperature_thresholds()[0][1]
         high = world.temperature_thresholds()[5][1]
         floor = 0
-        ceiling = 255
+        ceiling = 255  # could be changed into 16 Bit grayscale easily
 
         colors = numpy.interp(world.temperature['data'], [low, high], [floor, ceiling])
         colors = numpy.rint(colors).astype(dtype=numpy.int32)  # proper rounding
@@ -761,13 +729,13 @@ def draw_scatter_plot(world, size, target):
 
 
 def draw_simple_elevation_on_file(world, filename, sea_level):
-    img = ImagePixelSetter(world.width, world.height, filename)
+    img = PNGWriter.rgba_from_dimensions(world.width, world.height, filename)
     draw_simple_elevation(world, sea_level, img)
     img.complete()
 
 
 def draw_riversmap_on_file(world, filename):
-    img = ImagePixelSetter(world.width, world.height, filename)
+    img = PNGWriter.rgba_from_dimensions(world.width, world.height, filename)
     draw_riversmap(world, img)
     img.complete()
 
@@ -779,38 +747,38 @@ def draw_grayscale_heightmap_on_file(world, filename):
 
 
 def draw_elevation_on_file(world, filename, shadow=True):
-    img = ImagePixelSetter(world.width, world.height, filename)
+    img = PNGWriter.rgba_from_dimensions(world.width, world.height, filename)
     draw_elevation(world, shadow, img)
     img.complete()
 
 
 def draw_ocean_on_file(ocean, filename):
     height, width = ocean.shape
-    img = ImagePixelSetter(width, height, filename)
+    img = PNGWriter.rgba_from_dimensions(width, height, filename)
     draw_ocean(ocean, img)
     img.complete()
 
 
 def draw_precipitation_on_file(world, filename, black_and_white=False):
-    img = ImagePixelSetter(world.width, world.height, filename)
+    img = PNGWriter.rgba_from_dimensions(world.width, world.height, filename)
     draw_precipitation(world, img, black_and_white)
     img.complete()
 
 
 def draw_world_on_file(world, filename):
-    img = ImagePixelSetter(world.width, world.height, filename)
+    img = PNGWriter.rgba_from_dimensions(world.width, world.height, filename)
     draw_world(world, img)
     img.complete()
 
 
 def draw_temperature_levels_on_file(world, filename, black_and_white=False):
-    img = ImagePixelSetter(world.width, world.height, filename)
+    img = PNGWriter.rgba_from_dimensions(world.width, world.height, filename)
     draw_temperature_levels(world, img, black_and_white)
     img.complete()
 
 
 def draw_biome_on_file(world, filename):
-    img = ImagePixelSetter(world.width, world.height, filename)
+    img = PNGWriter.rgba_from_dimensions(world.width, world.height, filename)
     draw_biome(world, img)
     img.complete()
 
@@ -827,7 +795,7 @@ def draw_ancientmap_on_file(world, filename, resize_factor=1,
 
 
 def draw_scatter_plot_on_file(world, filename):
-    img = ImagePixelSetter(512, 512, filename)
+    img = PNGWriter.rgba_from_dimensions(512, 512, filename)
     draw_scatter_plot(world, 512, img)
     img.complete()
 
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -4,6 +4,7 @@ import numpy.ma as ma
 
 from worldengine.drawing_functions import draw_ancientmap, \
     draw_rivers_on_image
+from worldengine.image_io import PNGWriter
 
 # -------------
 # Helper values
@@ -772,8 +773,8 @@ def draw_riversmap_on_file(world, filename):
 
 
 def draw_grayscale_heightmap_on_file(world, filename):
-    img = ImagePixelSetter(world.width, world.height, filename)
-    draw_grayscale_heightmap(world, img)
+    img = PNGWriter.grayscale_from_array(world.elevation['data'], filename, scale_to_range=True)
+    #draw_grayscale_heightmap(world, img)
     img.complete()
 
 
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -801,6 +801,6 @@ def draw_scatter_plot_on_file(world, filename):
 
 
 def draw_satellite_on_file(world, filename):
-    img = ImagePixelSetter(world.width, world.height, filename)
+    img = PNGWriter(world.width, world.height, filename)
     draw_satellite(world, img)
     img.complete()
--- a/worldengine/drawing_functions.py
+++ b/worldengine/drawing_functions.py
@@ -40,11 +40,11 @@ def draw_rivers_on_image(world, target, factor=1):
 
     for y in range(world.height):
         for x in range(world.width):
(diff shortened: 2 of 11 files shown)


worldengine/draw.py
```diff
@@ -425,8 +395,8 @@ def draw_riversmap_on_file(world, filename):
 
 
 def draw_grayscale_heightmap_on_file(world, filename):
-    img = ImagePixelSetter(world.width, world.height, filename)
-    draw_grayscale_heightmap(world, img)
+    img = PNGWriter.grayscale_from_array(world.elevation['data'], filename, scale_to_range=True)
+    #draw_grayscale_heightmap(world, img)
```
tcld: I decided to just write out the array. I know this circumvents all modifications usually done in ```draw_grayscale_heightmap()``` but since this file is usable to import data into other programs I didn't like the idea of shifting around elevations.
If this is a bad idea, we can probably add in another parameter to decide which route to take - although I am not sure if a modified heightmap would prove useful at all.

worldengine/image_io.py
```diff
@@ -0,0 +1,215 @@
...
+  PNGReader.__init__()
+"""
+
+import numpy
+import png
+#Documentation PyPNG: https://pythonhosted.org/pypng/png.html
+#Documentation PurePNG: http://purepng.readthedocs.org/en/latest/
+#The latter one is a fork of the former one. It is yet to be seen which one is better.
+
+class PNGWriter(object):
```
tcld: Compared to the old ImageWriter-class this is restricted to PNG, but that one format it handles very well (due to PyPNG). I thought about adding in palette-support since most of worldengines output only uses a very small palette, but the class already seems like a bit of overkill and I didn't want to overcomplicate things.
If wanted, palette-support can go in, however.

ftomassetti:
Ok, so let's merge #161 first

worldengine/draw.py
tcld: When changing ```draw_simple_elevation()```, @esampson gave a reason for why it might be preferable to manipulate the elevation map.
Maybe you have a comment for this change?

ftomassetti: In general the idea is to have a function which do all the work on an "abstract image" (e.g., `draw_grayscale_heightmap`) and then a simple wrapper (`draw_grayscale_heightmap_on_file`). The reason is that the abstract image could either end up on an image file on disk OR it could be used to draw on the screen (as we do in worldengine-gui). This is the basic principle to follow to ensure that the user see on the screen what it will be later saved on disk.

If we need exceptions for performance reason I would just suggest to document that and keep them as exceptions.

tcld: PNGWriter can take any numpy-array and write it to a PNG-file. In the case of ```world.elevation['data']``` there is no immediate reason to perform any kind of processing before that. I don't see the need for any kind of change - so ```draw_grayscale_heightmap(world, img)``` could technically be an empty function.

This was not due to performance reasons, just because I wouldn't want to see the heightmap tampered with (since when using it as foundation for a 3D-landscape that would probably lead to visual irregularities).

tcld:
Ok, I think I understand what you mean. Since ```draw_grayscale_heightmap()``` is still around and in working order (although it will produce output that slightly differs from the raw elevation map) all of this should not cause problems right now.

Here it doesn't look like the functions in ```draw.py``` are actually used, though:
https://github.com/Mindwerks/worldengine-gui/blob/master/worldengine-gui/view.py

ftomassetti:
that could be because Bret reimplemented the GUI, we dropped the code I had from lands. I think in the long run we should use the same functions here and there but the GUI is still in alpha state for now I think. Probably @psi29a is the best person to comment on this

tcld:
Still, me not calling ```draw_grayscale_heightmap()``` in ```draw_grayscale_heightmap_on_file()``` would not stop the GUI from using ```draw_grayscale_heightmap()```. Right now it might just not be necessary to do that extra step in ```draw_grayscale_heightmap_on_file()```.

tcld:
About the three failing tests:
The grayscale-test obviously fails until a new (16 Bit) grayscale heightmap is pushed to worldengine-data.
The ancient-map tests might need a slightly extensive rewiring within ```drawing_functions.py```. I think I will take care of it later today, but there will be a lot of (very trivial) changes, as far as I can tell.

psi29a:
Couldn't this already be done via GDAL export? It was the reason why I dropped pypng support to begin with as GDAL can handle this for us.

tcld:
Yes, GDAL can do it. But creating a proper heightmap without GDAL seems like a good idea to me, the "normal" fake-grayscale heightmap isn't too useful anyway.
GDAL can still handle all possible other formats ( #159 ), including raw data.

tcld:
Additionally, Pillow can handle several different formats but doesn't fully support the PNG-standards. And since worldengine uses Pillow to exclusively write PNG-files, I thought that it might be a good idea to switch to a library that fully supports PNG, and *only* that.

psi29a:
It is a pity that 16-bit greyscale png isn't widely supported.

I know... been there with pypng as well. :)  

https://github.com/psi29a/worldsynth/blob/2b2e9f8d8b42f34f512b923a96e1fdc636f795b1/worldsynth.py#L635

tcld:
Well, I haven't found a program that couldn't at least open the output. *And* the file could be directly imported to Unreal Engine, which (I think) is a good example use-case for grayscale heightmaps.

Also, the class I wrote to handle PNG does support all formats (except the ones using palettes) so the output could be changed by switching just a small piece of code:
```python
img = PNGWriter.grayscale_from_array(world.elevation['data'], filename, scale_to_range=True)
```
becomes
```python
img = PNGWriter.grayscale_from_dimensions(world.width, world.height, filename, scale_to_range=True)
draw_grayscale_heightmap(world, img)
```

So at the very least nothing is lost.

tcld:
So, do you think I should pursue and finish this PR? Or do you really not like the idea of creating 16 Bit heightmaps by default?

psi29a:
Are you sure that gdal can't create 16-bit greyscale pngs? I thought I tested this, which is the reason I removed pypng since it would be redundant.

https://trac.osgeo.org/gdal/ticket/2806
^-- uint16 png works

tcld:
GDAL can do it for sure, I said that before. ;) But it is also a library that isn't too easy to install (for me, at least) and won't automatically come with the WorldEngine release. And I am not sure what uses the *fake* grayscale (it's actually RGBA made to look gray) 8 Bit heightmap as default output actually has.

I did design the PNGWriter-class to be a full replacement for the Pillow-writer - I didn't know that would kind of be a step backwards. I will think about this tomorrow.

tcld:
I decided to just finish this and let you decide. With leftover code from https://github.com/Mindwerks/worldengine/pull/141 there wasn't much work left to do, so I just went all the way - Pillow is gone, I even regenerated all the data (mostly because I needed the tests to hiccup in case x- and y-coordinates of arrays/images were accidentally swapped; since all test-maps were square before that was not assured). Python 2 and 3 passed all tests with that data on my system.

There are some more PyPNG abilities that could be taken advantage of, but I tried to leave output and code as unchanged as possible (there already are enough changes in this PR, even though most of them are tiny).

worldengine/drawing_functions.py
```diff
@@ -159,124 +159,124 @@ def _draw_cold_parklands(pixels, x, y):
     r = 105 - b
     g = 96 - b
     b = 38 - int(b / 2)
-    pixels[x, y] = (r, g, b, 255)
+    pixels[y, x] = (r, g, b, 255)
```
tcld: I started noticing in ```erosion.py``` that a couple of arrays were created by using ```(width, height)``` as dimensions - while all "major" arrays were created using ```(height, width)```. So I changed that and one thing lead to another...All changes in this file swap x- and y-coordinates. I ran a couple of tests to make sure the output stayed exactly the same, so there are some (~2) lines that do slightly more than just switching x and y.

worldengine/simulations/erosion.py
```diff
@@ -48,12 +48,12 @@ def is_applicable(self, world):
...
     def execute(self, world, seed):
-        water_flow = numpy.zeros((world.width, world.height))
-        water_path = numpy.zeros((world.width, world.height), dtype=int)
+        water_flow = numpy.zeros((world.height, world.width))
+        water_path = numpy.zeros((world.height, world.width), dtype=int)
         river_list = []
         lake_list = []
-        river_map = numpy.zeros((world.width, world.height))
-        lake_map = numpy.zeros((world.width, world.height))
+        river_map = numpy.zeros((world.height, world.width))
```
tcld: This is where the great xy-swapping of 2015 started.^^
I think this is an important change since it is confusing at best when in ```drawing_functions.py``` there can be code like this:
```python
if world.ocean[y, x]:
    world.river_map[x, y] = 5  # this is no ingenious trick; just different coordinate systems
```

psi29a:
I'm still curious as to what the problem was with GDAL on linux... it works here just with pip install pygdal (provided you have libgdal-dev package installed).

Any releases I upload (binary releases) will include gdal, that includes windows/osx/linux.

tcld:
It was pygdal that didn't work for me, I don't remember the exact reason, though - I switched PCs since so I don't have an immediate way to reproduce the problem either.

tcld:
A fresh environment:
```bash
(venv3)tcl@worldengine$ pip3 install pygdal
Collecting pygdal
  Using cached pygdal-1.11.2.1.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 20, in <module>
      File "/tmp/pip-build-4b773ez_/pygdal/setup.py", line 61
        print 'GDAL prefix from environment variable %s' % ENV_GDALHOME
                                                       ^
    SyntaxError: Missing parentheses in call to 'print'
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-4b773ez_/pygdal
```
I feel like I posted this before. pygdal really doesn't seem to be compatible with Python 3.

I also switched PCs again and still cannot get a proper Python 2 venv going. On the laptop it worked fine - but I will just go ahead and assume that pygdal will work fine with Python 2. It has to work *somewhere*, I guess.^^

On top of all those errors, I have the perfect solution for this PR - no need to include two ways to produce 16 Bit heightmaps, we can just remove ```draw_grayscale_heightmap()``` completely! :P (I really feel that the output isn't exactly useful in 8 Bit.)

psi29a:
Yeah, python2.  It was fixed here:
https://github.com/dezhin/pygdal/blob/master/setup.py#L61

just waiting for new release to drop.

tcld:
Oh ... I also feel like I read that before.^^

worldengine/draw.py
tcld: To sum this up and finish it:
The image for the grayscale-heightmap is now generated basically by *directly writing the corresponding numpy-array into an image-file*. There is no middleman necessary.

tcld:
I rebased #185 on this since before I wasn't able to have that PR pass it's new tests.

Since there are quite a lot of changes in here, I'll give a short rundown:
* The image-drawing class (PNGWriter) operates on a numpy-array and provides the same interface as the old ImagePixelSetter; pixels can be accessed/modified via the numpy-interface, if wanted.
* Pillow has been completely replaced by PyPNG.
* All access to the image-module happens inside *image_io.py*, the image-module could be swapped out easily.
* All direct access to data-carrying arrays (as well as images) happens via [y, x]; there were inconsistencies before where [x, y] was used instead. (Orientation of the images stays the same.)
* PNGReader reads PNGs and stores them in a numpy-array; the reader is used by the tests and later might want to be used in other parts of the code to circumvent the need for directly accessing the underlying image-module.
* At the moment RGBA-, RGB- and grayscale-PNGs can be created easily within a single line of code; palette-support was skipped for now.

psi29a:
Do we need Pillow at all anymore?  If we default to PyPNG (purely python) for all output, then use GDAL to support other things that Pillow normally supported, then we don't need Pillow.

tcld:
We don't need Pillow, I tried to remove it. Did I overlook something?

psi29a:
also in appveyor, tox, etc for the pillow->pypng switcheroo ?

tcld:
```tox.ini```, ```requirements.txt``` and ```setup.py``` should be taken care of. appveyor? I don't know what to change in there.

**Comment by psi29a** [comment]
This is super! :)

**Comment by psi29a** [comment]
Do we need new blessed images?

**Comment by tcld** [comment]
Some for sure. I don't remember which ones, though - I created a [full set](https://github.com/Mindwerks/worldengine-data/pull/11) of new worlds and images; non-square because I needed some feedback whether [y, x] etc. were misused.

**Comment by tcld** [comment]
This is still my next PR, some (if not all) of my current PRs are based on this. Maybe somebody finds some time to check the code; the x-y-replacements have been done using a regular expression, so there are not typos to be expected (and thus most of the changes in ```drawing_functions.py``` won't have to be reviewed in depth).

File: worldengine/drawing_functions.py [hunk 5]
```diff
@@ -502,13 +502,13 @@ def _dynamic_draw_a_mountain(pixels, rng, x, y, w=3, h=3):
...
-            pixels[x - itx, y + mody] = gradient(itx, darkarea, leftborder,
+            pixels[y + mody, x - itx] = gradient(itx, darkarea, leftborder,
                                                  (0, 0, 0), (64, 64, 64))
         for itx in range(-darkarea, lightarea + 1):
-            pixels[x + itx, y + mody] = gradient(itx, -darkarea, lightarea,
+            pixels[y + mody, x - itx] = gradient(itx, -darkarea, lightarea,
                                                  (64, 64, 64), (128, 128, 128))
         for itx in range(lightarea, leftborder):
-            pixels[x + itx, y + mody] = (181, 166, 127, 255)  # land_color
+            pixels[y + mody, x - itx] = (181, 166, 127, 255)  # land_color
```
tcld: First part of changes where I didn't just switch out x- and y-coordinate; I did some in-depth testing to get the exact same results as before.

psi29a: What do you want to do?

tcld: I think the image was rotated and mirrored due to the change in coordinates. That's why I added the minus in the x-component.

tcld:
I added a small comment to image_io.py regarding the [y, x]-notation, rebased the code to the current master and made some very small improvements. Is this ready to be merged yet?

Do you think that ```_dynamic_draw_a_mountain()``` should be removed? It isn't used yet and seemingly was never finished. Bot should esampson get done with his current work, the function would be already obsolete.

worldengine/drawing_functions.py
tcld: Should I maybe just remove the whole function? It is not used, not finished and (probably) soon will be obsolete due to https://github.com/Mindwerks/worldengine/issues/177.

tcld:
Data used to pass the tests: https://github.com/Mindwerks/worldengine-data/pull/11
I didn't *have* to regenerate all data (ancient maps, rivers and grayscale heightmap would have been enough) but I wanted to have non-square maps to make sure the changes to the xy-notation in this PR are alright.

psi29a:
The maps can be square and rectangular. That's fine. 

Do you consider this PR ready?

tcld:
I think it's ready, yes.

psi29a:
I don't see a PR for the WE-data?

tcld:
This one.
https://github.com/Mindwerks/worldengine-data/pull/11

psi29a:
Weird, I didn't get that an hour ago... I need sleep.

tcld:
Tests succeeded. :)
By the way: Will the Travis-tests be updated before a release? Or is that too much work?

psi29a:
What do you mean?

tcld:
All the tests that usually fail.

psi29a:
all the tests are green, otherwise it wouldn't say so.

do you mean environments?  python3, and the osx py2/py3?

tcld:
I meant all the optional tests that don't seem to work properly.
https://travis-ci.org/Mindwerks/worldengine/jobs/90341134
https://travis-ci.org/Mindwerks/worldengine/jobs/90341141
https://travis-ci.org/Mindwerks/worldengine/jobs/90341123

Do only I see those?

psi29a:
Yes, those are environments... not tests. And no, we can't fix that until Travis upgrades ubuntu 12.04 to something more recent and we have someone with osx experience to help get that running. Any volunteers? ;)

There are other factors to, pypy (on ubuntu 12.04)... we sit in a catch22 and I haven't figured out yet how to handle that.

tcld:
Ah, thanks for the explanation.

rbb:
What kind of experience/skills are needed for the OS X  environment testing?  I have an OS X box, and I run python on it via MacPorts. What else would I need?

psi29a:
Well, travis allows us to use osx in addition to linux (ubuntu 12.04).

I got travis to setup up the osx instances, but now we need to figure out how to actually get python (2.7 and 3.whatever) installed on them along with virtualenv and tox. Likely we can do that with brew/taps.

That would get us coverage on osx up and running. :)

psi29a:
btw.. merging this branch.

PR #167, opened 2015-10-31
fix loading pickle files

ftomassetti:
Apparently loading pickle files does not work anymore. Fixing.
Also adding a few spaces before the name of the operation, so that everything is aligned.

### DIFF
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -182,7 +182,7 @@ def __seems_protobuf_worldfile__(world_filename):
 
 def __seems_pickle_file__(world_filename):
     last_byte = __get_last_byte__(world_filename)
-    return last_byte == ord('.')
+    return str(last_byte) == '.'
 
 
 def load_world(world_filename):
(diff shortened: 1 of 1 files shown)


tcld:
It didn't work? At least one of the files used by the unit tests is a pickle file, and I never noticed any problems. Or was it just with user-loaded pickle files?

ftomassetti:
it was with user loaded pickle files: we have a function which look at a world file and try to guess if it is a pickle file or a protobuf file and depending on that load the world with a different function. The issue was with the function recognizing the type of file. Perhaps in the tests we call directly the method to load a file from pickle.

ftomassetti:
It is failing because of changes with the blessed images.

tcld:
You are right about the way the pickle files are loaded in the tests.

I wouldn't mind if you just merged this, if that isn't against standard procedure. The change is kind of small.^^

ftomassetti:
#168 was built on top of this so no need to merge this one separately

Issue #119, opened 2015-10-03
Problem with Rivers


esampson:
No one seems to have brought this up before but there seems to be a problem with river generation. For some reason they have an extremely strong tendency to run to the northwest.
Here is a river map generated from the latest build and I can't see a  single river that runs outside of the range of 270-360 degrees north. This also doesn't seem to be a new occurrence. It is something I noticed a while back but I just didn't get around to filing an issue on it.
![seed_28427_rivers](https://cloud.githubusercontent.com/assets/8743461/10260664/39e8c368-692f-11e5-81f7-4d8c24ddbae7.png)

psi29a:
That is because the current algorithm uses wind and rain maps, the water collects above sea-level and is blown in just one direction. As precipitation falls, then further along in that direction is less rain to fall. 

In my original code, the wind direction was random... so I'm going to assume that it is the same as before. This means that there will be areas with no rain at all which contributes to biomes.

You can verify this by looking at your biome map, if there are deserts or similar biomes in the area, then it is likely that there isn't any rainfall and rivers there. Rivers can begin in other biomes and flow into deserts however... just like the Nile.

esampson:
Are rivers really that affected by wind in the real world?

psi29a:
Interesting that you have asked that, if there is no wind to carry precipitation, then you'll never get rain or snow in some areas. That means no snow-pack that melts, no erosion through run-off, no collection into into streams that later collect into other streams than into rivers... 

Another effect is rain-shadows, that when wind hits mountains that precipitation falls heavily on that side of the mountain. The result is that on the other side gets little to no rain-fall as a result.

The only thing you'll have are brons or sources, that bubble up from out the ground... that is based on aquifers and underground rivers.

There is a bit of overlap here, but these two things can be considered separate.

esampson:
Ok, I realize that the direction of the wind is going to affect rainfall, but it doesn't seem like it should have that strong an effect on the rivers themselves. Once a river has formed it always has to run downhill, but in the example above (seed 28427) I can find at least one spot where it seems to clearly be running up hill (coordinates 338, 172) by laying the river map over the grayscale heightmap.
![seed_28427_grayscale_zoom](https://cloud.githubusercontent.com/assets/8743461/10291875/f073580e-6b60-11e5-9b1b-75e384464eb4.png)
[...]

PR #168, opened 2015-10-31
Rivers worldsynth

ftomassetti:
We already calculated rivermap and lakemap but then we used an ad-hoc routine to draw rivers, with questionable results.

This patch is built on top of #167 

See images in #119

### DIFF
--- a/worldengine/drawing_functions.py
+++ b/worldengine/drawing_functions.py
@@ -39,50 +39,16 @@ def draw_rivers_on_image(world, target, factor=1):
     """Draw only the rivers, it expect the background to be in place
     """
 
-    def _lowest_neighbour(world, pos):
-        x, y = pos
-        lowest = None
-        lowest_lvl = None
-        for dx in range(-1, 1):
-            for dy in range(-1, 1):
-                if dx != 0 or dy != 0:
-                    e = world.elevation['data'][y + dy, 
-                        x + dx]  # +world.humidity['data'][y+dy, x+dx]/3.0
-                    if (not lowest_lvl) or (e < lowest_lvl):
-                        lowest_lvl = e
-                        lowest = (x + dx, y + dy)
-                    elif (e == lowest_lvl) and ((x + y + dx + dy) % 2 == 0):
-                        lowest_lvl = e
-                        lowest = (x + dx, y + dy)
-
-        return lowest
-
-    def _draw_river(world, target, pos, factor):
-        if world.ocean[pos[1], pos[0]]:
-            return
-        x, y = pos
-        for dx in range(factor):
-            for dy in range(factor):
-                target.set_pixel(x * factor + dx, y * factor + dy,
-                                 (0, 0, 128, 255))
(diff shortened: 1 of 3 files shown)


worldengine/drawing_functions.py
```diff
@@ -40,50 +40,16 @@ def draw_rivers_on_image(world, target, factor=1):
...
-        if world.ocean[pos[1], pos[0]]:
-            return
-        x, y = pos
-        for dx in range(factor):
-            for dy in range(factor):
-                target.set_pixel(x * factor + dx, y * factor + dy,
-                                 (0, 0, 128, 255))
-        _draw_river(world, target, _lowest_neighbour(world, pos), factor)
-
-    n_rivers = int(math.sqrt(world.width * world.height))
```
tcld: Now the math module doesn't need to be imported anymore.

ftomassetti: Removed, thanks

worldengine/drawing_functions.py
```diff
@@ -40,50 +40,16 @@ def draw_rivers_on_image(world, target, factor=1):
...
-    for i in range(1, n_rivers):
-        candidates = []
-        for j in range(1, 10):
-            candidates.append(_pseudo_random_land_pos(world, i * j + j))
```
tcld: I think that _pseudo_random_land_pos() is now not used anymore, either. I would like to see it removed.^^

worldengine/drawing_functions.py
```diff
@@ -40,50 +40,16 @@ def draw_rivers_on_image(world, target, factor=1):
...
+    for y in range(world.height):
+        for x in range(world.width):
+            if world.is_land((x, y)) and (world.river_map[x, y] > 0.0):
+                for dx in range(factor):
+                    for dy in range(factor):
+                        target.set_pixel(x * factor + dx, y * factor + dy, (0, 0, 128, 255))
+            if world.is_land((x, y)) and (world.lake_map[x, y] != 0):
+                for dx in range(factor):
+                    for dy in range(factor):
+                        target.set_pixel(x * factor + dx, y * factor + dy, (0, 100, 128, 255))
```
tcld: Simple and effective code, I don't see anything that could go wrong. Certainly a huge improvement over the kind of complicated previous code. (And obviously the results are a lot better, too. :) )

worldengine/drawing_functions.py
ftomassetti: correct, removing

tcld:
This will improve the results of world generation *so much*!

The code reminds me a little bit of what ```irrigation.py``` looked like before I replaced its core with numpy code - I guess that drawing the rivers and lakes is kind of a slow process? I wouldn't know how to improve it, I'm just curious.

psi29a:
*thumbs up* just need new blessed images?

ftomassetti:
it seems so. I will commit the new blessed images and merge then.

ftomassetti:
Blessed images updated, tests restarted

ftomassetti:
rebasing on master and then restarting tests

PR #170, opened 2015-11-01
Update README.md

esampson:


README.md
```diff
@@ -148,73 +148,29 @@ _Note that options were changed in version 0.5.3_
 For example these commands:
 
 ```python
-worldengine world -s 4 -n an_example -p 2048 -q 25 -x 2048 -y 2048
+worldengine -b -s 4 -n an_example -q 25 -x 2048 -y 2048
```
ftomassetti: ok for removing `world` but perhaps we could avoid using `-b` here because it just changes the format we use for saving the world, so there is no visible effect here

README.md
```diff
@@ -148,73 +148,29 @@ _Note that options were changed in version 0.5.3_
...
 -----------------------
- seed              : 4
- name              : seed3
- width             : 2048
- height            : 2048
- plates resolution : 2048
- number of plates  : 25
- operation         : world generation
- step              : full
+ operation              : world generation
```
ftomassetti: perhaps we could display the output as it was correctly aligned, I will send a PR to correctly align it

README.md
```diff
@@ -148,73 +148,29 @@ _Note that options were changed in version 0.5.3_
...
+ world format         : protobuf
+ black and white maps : False
+ step                 : full
+ greyscale heightmap  : False
+ rivers map           : False
+ scatter plot         : False
+ fade borders         : True
 
 starting (it could take a few minutes) ...
-...plates simulated
```
ftomassetti: I think these lines are still produced. Do you think we should remove them? In that case we could remove also the line above (starting...)

README.md
```diff
@@ -225,6 +181,7 @@ Producing ouput:
 * elevation image generated in './an_example_elevation.png'
 ...done
 
+
```
ftomassetti: Is this extra white line intended?

README.md
tcld: Having an exact copy of the verbose-output in the README is kind of ugly when it comes to changing or adding messages (the README will probably be overlooked). I think this should be left out - as an alternative an image of the final biomes would convey roughly the same information and provide some eye-candy on top of it.

esampson:
Made suggested changes. No clue why this would cause a coveralls decrease since it doesn't affect any of that.

tcld:
I was told that coveralls is a bit random - sometimes the result changes between basically empty commits.

I hope the others are fine with the removal of the exemplary output, since there was no further comment about it. (I like it, of course.^^)

esampson:
It actually isn't all that random, as I understand it. I mean, sure, if all you do is run the program once and then see what lines of code run and what don't it can be a bit random since not every biome will be generated. We take a lot of the randomness out of it, however, by doing more comprehensive tests that do things such as force all biomes to be evaluated.

tcld:
By the way: I noticed that you (?) added a biome testworld to worldengine-data. I don't know what's in there, but should it have to be regenerated at some point...how would one do it? You might want to add some sort of instruction (or add a line to data_generator.py).

esampson:
Yes, that was me and no, it shouldn't need to be regenerated any time in the foreseeable future. It is a small world (I think it is 256x256) that was generated with an increasing temperature going from west to east and increasing humidity going from north to south (or something very similar. I may have the directions wrong). 

As a result the world has all possible biomes while maintaining a small size. This lets tests be run with the certainty that every biome is encountered so that we know that there's no errors in any of that code (and it also decreases the randomness of the coverage test).

esampson:
Oh, the world also has no ocean so there would be no danger that one of the biomes that needed to be tested was underwater.

ftomassetti:
LGTM, thanks!

PR #171, opened 2015-11-02
Added command line options for temperature/precipitation gamma correction curve.

esampson:
Also minor code cleanup.

Modified scatter plot to show gamma correction curve.

### DIFF
--- a/tests/serialization_test.py
+++ b/tests/serialization_test.py
@@ -11,7 +11,7 @@ class TestSerialization(unittest.TestCase):
         self.maxDiff = None
 
     def test_pickle_serialize_unserialize(self):
-        w = world_gen("Dummy", 32, 16, 1, [.874, .765, .594, .439, .366, .124], [.941, .778, .507, .236, 0.073, .014, .002], step=Step.get_by_name("full"))
+        w = world_gen("Dummy", 32, 16, 1, step=Step.get_by_name("full"))
         f = tempfile.NamedTemporaryFile(delete=False).name
         w.to_pickle_file(f)
         unserialized = World.from_pickle_file(f)
@@ -37,7 +37,7 @@ class TestSerialization(unittest.TestCase):
         self.assertEqual(w, unserialized)
 
     def test_protobuf_serialize_unserialize(self):
-        w = world_gen("Dummy", 32, 16, 1, [.874, .765, .594, .439, .366, .124], [.941, .778, .507, .236, 0.073, .014, .002], step=Step.get_by_name("full"))
+        w = world_gen("Dummy", 32, 16, 1, step=Step.get_by_name("full"))
         serialized = w.protobuf_serialize()
         unserialized = World.protobuf_unserialize(serialized)
         self.assertTrue(_equal(w.elevation['data'], unserialized.elevation['data']))
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -22,11 +22,12 @@ STEPS = 'plates|precipitations|full'
 
 
 def generate_world(world_name, width, height, seed, num_plates, output_dir,
-                   step, ocean_level, temps, humids,
-                   world_format='pickle', fade_borders=True,
+                   step, ocean_level, temps, humids, world_format='pickle',
+                   gamma_curve=1.25, curve_offset=.2, fade_borders=True,
                    verbose=True, black_and_white=False):
     w = world_gen(world_name, width, height, seed, temps, humids, num_plates, ocean_level,
-                  step, fade_borders=fade_borders, verbose=verbose)
+                  step, gamma_curve=gamma_curve, curve_offset=curve_offset,
+                  fade_borders=fade_borders, verbose=verbose)
 
     print('')  # empty line
     print('Producing ouput:')
@@ -313,6 +314,14 @@ def main():
                              "If not provided, the default values will be used. \n" +
                             "[default = .059/.222/.493/.764/.927/.986/.998]",
                             metavar="#/#/#/#/#/#/#")
+    g_generate.add_argument('-gv', '--gamma-value', dest='gv', type=float,
+                        help="N = Gamma value for temperature/precipitation " +
+                             "gamma correction curve. [default = %(default)s]",
+                        metavar="N", default='1.25')
+    g_generate.add_argument('-go', '--gamma-offset', dest='go', type=float,
+                        help="N = Adjustment value for temperature/precipitation " +
+                             "gamma correction curve. [default = %(default)s]",
+                        metavar="N", default='.2')
     g_generate.add_argument('--not-fade-borders', dest='fade_borders', action="store_false",
                                help="Not fade borders",
                                default=True)
@@ -441,10 +450,10 @@ def main():
     if args.temps and len(args.temps.split('/')) is not 6:
         usage(error="temps must have exactly 6 values")
 
-    temps =[.874, .765, .594, .439, .366, .124]
+    temps = [.874, .765, .594, .439, .366, .124]
     if args.temps:
         temps = args.temps.split('/')
-        for x in range(0,6):
+        for x in range(0,len(temps)):
             temps[x] = 1 - float(temps[x])
 
     if args.humids and not generation_operation:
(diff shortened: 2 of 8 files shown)


worldengine/cli/main.py
```diff
@@ -313,6 +314,14 @@ def main():
                              "If not provided, the default values will be used. \n" +
                             "[default = .059/.222/.493/.764/.927/.986/.998]",
                             metavar="#/#/#/#/#/#/#")
+    g_generate.add_argument('-gv', '--gamma-value', dest='gv', type=float,
```
ftomassetti: We really need good documentation for these values :)

worldengine/cli/main.py
```diff
@@ -441,10 +450,10 @@ def main():
     if args.temps and len(args.temps.split('/')) is not 6:
         usage(error="temps must have exactly 6 values")
 
-    temps =[.874, .765, .594, .439, .366, .124]
+    temps = [.874, .765, .594, .439, .366, .124]
     if args.temps:
         temps = args.temps.split('/')
-        for x in range(0,6):
+        for x in range(0,len(temps)):
```
ftomassetti: should we verify the input is correct? (correct number of items, values in correct range, decreasing order)

worldengine/world.py
```diff
@@ -34,6 +35,8 @@ def __init__(self, name, width, height, seed, num_plates, ocean_level,
         self.step = step
         self.temps = temps
         self.humids = humids
+        self.gamma_curve = gamma_curve
```
ftomassetti: we should serialize these values. I opened an issue for that (#174) as we can do that also after having merged this patch

worldengine/cli/main.py
tcld: Yes. When I hear "gamma value" the first thing I think of is something related to the brightness of the image-output.
I would maybe like to see a link to some page that explains this in more detail.

tests/serialization_test.py
```diff
@@ -11,7 +11,7 @@ def setUp(self):
         self.maxDiff = None
 
     def test_pickle_serialize_unserialize(self):
-        w = world_gen("Dummy", 32, 16, 1, [.874, .765, .594, .439, .366, .124], [.941, .778, .507, .236, 0.073, .014, .002], step=Step.get_by_name("full"))
+        w = world_gen("Dummy", 32, 16, 1, step=Step.get_by_name("full"))
```
tcld: I like this a lot better. :)

worldengine/cli/main.py
```diff
@@ -313,6 +314,14 @@ def main():
                              "If not provided, the default values will be used. \n" +
                             "[default = .059/.222/.493/.764/.927/.986/.998]",
                             metavar="#/#/#/#/#/#/#")
+    g_generate.add_argument('-gv', '--gamma-value', dest='gv', type=float,
+                        help="N = Gamma value for temperature/precipitation " +
```
tcld: What is ```N = ``` good for?

worldengine/cli/main.py
```diff
@@ -313,6 +314,14 @@ def main():
...
+                             "gamma correction curve. [default = %(default)s]",
+                        metavar="N", default='1.25')
```
tcld: Do you actively use the ```metavar```-name somewher?
(Also in line 324.)

As far as I understand (https://docs.python.org/3/library/argparse.html#metavar) you shouldn't make use of that parameter and the (much more useful) ```dest```-name will be used in case it is needed.

worldengine/cli/main.py
tcld: I would suggest to add an assert to test the length of the array (otherwise a user might see crashes much later in the code and not understand them). Regarding the value-order etc. I would just leave that to the responsibility of the user. Should we start working on a global settings-file in the future, we could add a class to handle all kinds of stupid inputs.

(This also affects line 468.)

ftomassetti: I think that also issues in the order or the value (range 0..1) could lead to crashes or at least unpredictable behavior later. The best person for confirming/denying this is probably @esampson 

I do not think we should use an assert but instead handle the errors printing a nice message and closing the application with an error code.

worldengine/cli/main.py
esampson: That's because this is the same formula.

Y = X^gamma where both X and Y are in the range of 0-1.

worldengine/cli/main.py
esampson: We do test the length of the array. We don't test order because there is a very slight possibility someone will want to input a series where parameter X is less than parameter X-1 (this would result in a zone being skipped).

worldengine/cli/main.py
esampson: Metavar regulates how the option is displayed when you use -help.

worldengine/cli/main.py
ftomassetti: mmm I see. I would in that case issue a warning, so that users who invert the order of items by mistake could realize it. If then they would like to do so... free to do it

tcld:
I noticed something:
In the current master-branch https://github.com/Mindwerks/worldengine/blob/master/tests/data/data_generator.py is broken due to missing parameters. You probably overlooked that file in the PR preceding this one. It would be nice if you could double-check that it works with this PR. (And it would be nice if this PR then could be merged as soon as possible. ;) )

esampson:
Added a routine to print warnings. Added warnings for unusual but legal assignments for humidity and temperature ranges. Added error catching for gamma value and gamma offset.

tcld:
Good job on that front. :) I hope that we can at some point tackle "global variables" in a way that shortens that code a bit (or rather puts it somewhere else^^).

(Also, please don't overlook my last comment.)

esampson:
Ok. Fixed data_generator.py by adding defaults to _plates_simulation. Also added a coverage test for the new warnings systems.

worldengine/cli/main.py
```diff
@@ -482,6 +500,26 @@ def main():
         print(' draw rivers            : %s' % args.draw_rivers)
         print(' draw mountains         : %s' % args.draw_mountains)
         print(' draw land outer border : %s' % args.draw_outer_border)
+        
+    #Warning messages
+    warnings = []
+    if temps != sorted(temps, reverse=True):
+        warnings.append("WARNING: Temperature array not in ascending order")
+    if numpy.amin(temps) < 0:
+        warnings.append("WARNING: Maximum value in temperature array greater than 1")
```
ftomassetti: perhaps the messages are inverted

worldengine/plates.py
```diff
@@ -36,25 +36,30 @@ def generate_plates_simulation(seed, width, height, sea_level=0.65,
...
+                  humids, gamma_curve, curve_offset)
     world.set_elevation(numpy.array(e_as_array).reshape(height, width), None)
     world.set_plates(array_to_matrix(p_as_array, width, height))
     return world
 
 
-def world_gen(name, width, height, seed, temps, humids, num_plates=10, ocean_level=1.0,
-              step=Step.full(), fade_borders=True, verbose=get_verbose()):
+def world_gen(name, width, height, seed, temps=[.874, .765, .594, .439, .366, .124],
+              humids=[.941, .778, .507, .236, 0.073, .014, .002], num_plates=10,
```
ftomassetti: could we use a constant to avoid repeating the default values for humids and temps?

ftomassetti:
Thanks for adding the warnings!

ftomassetti:
No idea why AppVeyor failed. Restarting it.

ftomassetti:
Apparently the last build was killed because it was taking too long (and it was taking too long during the setup phase). I do not see a way to restart the build.

psi29a:
Bumped it for a rebuild. First log-in, then you can get the rebuild button. :)

ftomassetti:
All the tests passing. Merging. Thanks to everyone!

worldengine/cli/main.py
esampson: No, the warnings are actually correct. We invert the values that are given because I think it will be easier for people to understand and I worry that if I go through all of the code to handle the new values directly I'll end up missing bits and it will end up not working correctly.

PR #173, opened 2015-11-02
realign operation

ftomassetti:


psi29a:
I like stable coverage.... :)

Issue #166, opened 2015-10-31
Improving documentation


ftomassetti:
In this issue we could discuss how to improve documentation. I think it could be also a good exercise to take a step back and look at the global picture so that we can shape how we want version 1.0 to be.

I would suggest to use readthedocs: https://readthedocs.org/. To use it we should create a directory containing documentation in a rst format (think of it as an alternative to Markdown) and a git hook will just take those documents, compile them and produce documentation which would be available under worldengine.readthedocs.org

What do you think? @psi29a @esampson @tcld

tcld:
I don't have experience with writing proper documentation. Are you thinking of a user-side documentation (explain commandline parameters and maybe some scientific foundations) or more of a developer-side documentation (explain functions, return values, etc.)?

ftomassetti:
User-side documentation: what it is possible to do with WorldEngine. We could need also better developer-side documentation. Perhaps we could use comments in code and the wiki for that.

ftomassetti:
I woud like to document:
* the single options showing different worlds generated changing the parameters
* typical use cases
  * I want to use WorldEngine as a library from Python
  * I want to use WorldEngine as a library from Java
  * I want to just generate large cool maps
  * I want to import some map and run some simulations in WorldEngine

esampson:
I can look at tackling the general documentation; things such as what the different options are, what they produce, and explanations of biomes, temperature and humidity ranges.

ftomassetti:
thanks @esampson that is appreciated. Perhaps we could try to have small PRs for the documentation and merge frequently so it is easier for everyone to contribute. Do you want me to setup readthedocs or do you want to do that?

psi29a:
RTD is great, but we should really use Sphinx to handle things. I've experience with this, just a bit to setup and it documents itself. :)

psi29a:
This means that documentation needs to be written in ReST:
http://sphinx-doc.org/rest.html

If people want to start doing this, it is trivial to add to Sphinx. RTD then takes Sphinx and generates documentation for us in LaTeX, pdf, HTML and etc...

ftomassetti:
Cool. We have the manual in master now so we can discuss the content in specific issues. Closing this general one.

PR #179, opened 2015-11-03
manual: setup

ftomassetti:
I just setup the manual on readthedocs. It is visible here: http://worldengine.readthedocs.org/en/manual/

We could use this for user documentation. What do you think?

See #166


esampson:
While I like online docs I am also a firm believer in supplying offline versions. Nothing worse than wanting to use a piece of software that doesn't require an internet connection but not being able to because you need to look something up and the manual does require one (which you currently lack, for some reason).

ftomassetti:
You can also build the manual in several formats. We could generate it make it available either with the program or as a separate download.

psi29a:
This is a nice start, but keep Sphinx in mind please. :)

psi29a:
Nice start. :)

ftomassetti:
stupid question: isn't sphinx what is used by readthedocs?

ftomassetti:
I confirm that we can generate from the same format this:
```
Please use `make <target>' where <target> is one of
  html       to make standalone HTML files
  dirhtml    to make HTML files named index.html in directories
  singlehtml to make a single large HTML file
  pickle     to make pickle files
  json       to make JSON files
  htmlhelp   to make HTML files and a HTML help project
  qthelp     to make HTML files and a qthelp project
  applehelp  to make an Apple Help Book
  devhelp    to make HTML files and a Devhelp project
  epub       to make an epub
  latex      to make LaTeX files, you can set PAPER=a4 or PAPER=letter
  latexpdf   to make LaTeX files and run them through pdflatex
  latexpdfja to make LaTeX files and run them through platex/dvipdfmx
  text       to make text files
  man        to make manual pages
  texinfo    to make Texinfo files
  info       to make Texinfo files and run them through makeinfo
  gettext    to make PO message catalogs
  changes    to make an overview of all changed/added/deprecated items
  xml        to make Docutils-native XML files
  pseudoxml  to make pseudoxml-XML files for display purposes
  linkcheck  to check all external links for integrity
  doctest    to run all doctests embedded in the documentation (if enabled)
  coverage   to run coverage check of the documentation (if enabled)
```

So we should be able to provide offline manuals

ftomassetti:
Let's merge this so that everyone is able to improve it

psi29a:
Yes, RTD uses sphinx and our sphinx/config/python/rst files to generate the documents.

tcld:
I just noticed that there are problems with the [manifest-test](https://travis-ci.org/Mindwerks/worldengine/jobs/89637800) in recent PRs. Seemingly introduced with #179 . I seem to always forget to check the optional tests. :(

ftomassetti:
Which problems?

tcld:
Rules need to be added to ```MANIFEST.in```, as suggested in the link I gave.

psi29a:
Since documents (end result) is stored in RTD, we don't need to ship anything having to do with docs.

PR #181, opened 2015-11-04
Satellite map (non-messed-up version)

pangal:
Here is a pull request with the proper issues sorted out.

This branch adds functionality for creating a "Satellite map" of the generated world. This is currently an option, default false, toggled on with the --sat command when running.

Additional work may need to be done in selecting more appropriate biome colors and tuning the algorithm to display elevation / shading, to get a more realistic result.

Some sample outputs are attached here:
8e13c412-819c-11e5-938f-a05a3379e9de
8e177a62-819c-11e5-8452-9a03a9fa8734

### DIFF
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -7,7 +7,7 @@ import worldengine.generation as geo
 from worldengine.common import array_to_matrix, set_verbose, print_verbose
 from worldengine.draw import draw_ancientmap_on_file, draw_biome_on_file, draw_ocean_on_file, \
     draw_precipitation_on_file, draw_grayscale_heightmap_on_file, draw_simple_elevation_on_file, \
-    draw_temperature_levels_on_file, draw_riversmap_on_file, draw_scatter_plot_on_file
+    draw_temperature_levels_on_file, draw_riversmap_on_file, draw_scatter_plot_on_file, draw_satellite_on_file
 from worldengine.plates import world_gen, generate_plates_simulation
 from worldengine.imex import export
 from worldengine.step import Step
@@ -82,6 +82,9 @@ def draw_scatter_plot(world, filename):
     draw_scatter_plot_on_file(world, filename)
     print("+ scatter plot generated in '%s'" % filename)
 
+def draw_satellite_map(world, filename):
+    draw_satellite_on_file(world, filename)
+    print("+ satellite map generated in '%s'" % filename)
 
 def generate_plates(seed, world_name, output_dir, width, height,
                     num_plates=10):
@@ -317,6 +320,9 @@ def main():
     g_generate.add_argument('--scatter', dest='scatter_plot',
                             action="store_true", help="generate scatter plot")
 
+    g_generate.add_argument('--sat', dest='satelite_map',
+                            action="store_true", help="generate satellite map")
+
     # -----------------------------------------------------
     g_ancient_map = parser.add_argument_group(
         "Ancient Map Options", "These options are only useful in " +
@@ -469,6 +475,7 @@ def main():
         print(' greyscale heightmap  : %s' % args.grayscale_heightmap)
         print(' rivers map           : %s' % args.rivers_map)
         print(' scatter plot         : %s' % args.scatter_plot)
+        print(' satellite map        : %s' % args.satelite_map)
         print(' fade borders         : %s' % args.fade_borders)
     if args.temps:
         print(' temperature ranges   : %s' % args.temps)
@@ -503,6 +510,9 @@ def main():
         if args.scatter_plot:
             draw_scatter_plot(world,
             '%s/%s_scatter.png' % (args.output_dir, world_name))    
+        if args.satelite_map:
+            draw_satellite_map(world,
+            '%s/%s_satellite.png' % (args.output_dir, world_name))    
 
     elif operation == 'plates':
         print('')  # empty line
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -10,7 +10,7 @@ from worldengine.drawing_functions import draw_ancientmap, \
 # -------------
 
 ### For draw_satellite ###
-NOISE_RANGE = 10 # a random value between -NOISE_RANGE and NOISE_RANGE will be added to the rgb of each pixel
+NOISE_RANGE = 15 # a random value between -NOISE_RANGE and NOISE_RANGE will be added to the rgb of each pixel
 
 # These are arbitrarily-chosen elevation cutoffs for 4 different height levels. 
 # Some color modifiers will be applied at each level
@@ -24,11 +24,18 @@ HILL_ELEV          = 145
 HIGH_MOUNTAIN_NOISE_MODIFIER = (10, 6,   10)
 MOUNTAIN_NOISE_MODIFIER =      (-4, -12, -4)
 HIGH_HILL_NOISE_MODIFIER =     (-3, -10, -3)
-HILL_NOISE_MODIFIER =          (-2, -6, -2)
+HILL_NOISE_MODIFIER =          (-2, -6,  -2)
 
 # This is the base "mountain color". Elevations above this size will have their colors interpolated with this 
 # color in order to give a more mountainous appearance
 MOUNTAIN_COLOR = (50, 57, 28)
+
+# If a tile is a river or a lake, the color of the tile will change by this amount
+RIVER_COLOR_CHANGE = (-12, -12, 4)
+LAKE_COLOR_CHANGE = (-12, -12, 10)
+
+# The normalized (0-255) value of an elevation of a tile gets divided by this amount, and added to a tile's color
+BASE_ELEVATION_INTENSITY_MODIFIER = 10
 
 # How many tiles to average together when comparing this tile's elevation to the previous tiles.
 SAT_SHADOW_SIZE = 5
@@ -208,10 +215,22 @@ def _sature_color(color):
 def elevation_color(elevation, sea_level=1.0):
     return _sature_color(_elevation_color(elevation, sea_level))
 
+def add_colors(*args):
+    ''' Do some *args magic to return a tuple, which has the sums of all tuples in *args '''
+    # Adapted from an answer here: http://stackoverflow.com/questions/14180866/sum-each-value-in-a-list-of-tuples
+    return tuple([sum(x) for x in zip(*args)])
 
-def get_normalized_elevation_mask(world):
+def average_colors(c1, c2):
+    ''' Average the values of two colors together '''
+    r = int((c1[0] + c2[0])/2)
+    g = int((c1[1] + c2[1])/2)
+    b = int((c1[2] + c2[2])/2)
+
+    return (r, g, b)
+
+def get_normalized_elevation_array(world):
     ''' Convert raw elevation into normalized values between 0 and 255,
-        and return a numpy arrawy of these values '''
+        and return a numpy array of these values '''
 
     e = world.elevation['data']
 
@@ -249,7 +268,7 @@ def get_biome_color_based_on_elevation(world, elev, x, y):
         Finally, the noise plus the biome color are added and returned
     '''
     v = world.biome[y, x]
-    biome_r, biome_g, biome_b = _biome_satellite_colors[v]
+    biome_color = _biome_satellite_colors[v]
 
     # Default is no noise - will be overwritten if this tile is land
     noise = (0, 0, 0)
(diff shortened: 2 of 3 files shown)


worldengine/generation.py
```diff
@@ -200,7 +200,7 @@ def generate_world(w, step):
 
     # Prepare sufficient seeds for the different steps of the generation
     rng = numpy.random.RandomState(w.seed)  # create a fresh RNG in case the global RNG is compromised (i.e. has been queried an indefinite amount of times before generate_world() was called)
-    sub_seeds = rng.randint(0, 4294967295, size=100)  # sys.maxsize didn't quite work
+    sub_seeds = rng.randint(0, 2147483647, size=100)  # sys.maxsize didn't quite work
```
ftomassetti: if the larger value cause an issue on win32 I would be ok with changing it. @tcld is that fine for you?

worldengine/draw.py
```diff
@@ -209,28 +393,95 @@ def draw_riversmap(world, target):
...
+                if all_r:
+                    avg_r = int(sum(all_r) / len(all_r))
+                    avg_g = int(sum(all_g) / len(all_g))
+                    avg_b = int(sum(all_b) / len(all_b))
+
+                    ## Setting color of the pixel again - this will be once more modified by the shading algorithm
+                    target.set_pixel(x, y, (avg_r, avg_g, avg_b, 255))
+
+    # "Shade" the map by sending beams of light west to east, and increasing or decreasing value of pixel based on elevation difference
+    for y in xrange(SAT_SHADOW_SIZE-1, world.height-SAT_SHADOW_SIZE-1):
```
ftomassetti: We tend to use range instead of xrange because xrange is not available on Python 3

ftomassetti:
Thank you for going through the effort of re-preparing the pull-request: it is appreciated. Git drove all of us mad at the beginning (honestly, it still does from time to time...)

worldengine/generation.py
pangal: I don't know if it's win32 - I'm on a 64-bit system but have a 32-bit python install, and that was causing issues for me.

worldengine/draw.py
pangal: Ah, I was wondering about that.

worldengine/generation.py
psi29a: Please use `six.MAXSIZE` which will work on python2 and python3.

tcld: I didn't know the upper limit could differ. 
I would suggest sticking to the new fixed number and add a new comment like "max 2^31-1 due to... "
That way worlds generated under Python 2 and 3 won't be different, which six won't guarantee.

psi29a: While I agree with you @tcld, :(

<3 six

psi29a: Again, best to use six.MAXINT

tcld: numpy.randint handles uint32 at most. Even if that differed for different systems, we have to stick to the smallest common denominator, which seems to be int32 now.

tcld: I would prefer some kind of constant, though. I just didn't find any.

worldengine/cli/main.py
```diff
@@ -317,6 +320,9 @@ def main():
     g_generate.add_argument('--scatter', dest='scatter_plot',
                             action="store_true", help="generate scatter plot")
 
+    g_generate.add_argument('--sat', dest='satelite_map',
```
tcld: I wouldn't mind making this a default output. It gives a beautiful overview over the world.
What do you think, @ftomassetti and @psi29a ?

worldengine/draw.py
```diff
@@ -1,6 +1,7 @@
 from PIL import Image
 import numpy
 import numpy.ma as ma
+import random
```
tcld: Don't use random, use numpy.randint() etc. instead.

worldengine/draw.py
```diff
@@ -9,6 +10,36 @@
 # Helper values
 # -------------
 
+### For draw_satellite ###
+NOISE_RANGE = 10 # a random value between -NOISE_RANGE and NOISE_RANGE will be added to the rgb of each pixel
+
+# These are arbitrarily-chosen elevation cutoffs for 4 different height levels. 
+# Some color modifiers will be applied at each level
+HIGH_MOUNTAIN_ELEV = 215
```
tcld: In other places of the code relative values are used, e.g. "highest 10% of the map". Making use of find_threshold() in basic_operations.py would then lead to the corresponding absolute value.
The used values are stored somewhere, I believe. Something like world.elevation['thresholds'].

I can't verify that myself at the moment, maybe you can take a look.

worldengine/draw.py
```diff
@@ -129,6 +210,109 @@ def elevation_color(elevation, sea_level=1.0):
     return _sature_color(_elevation_color(elevation, sea_level))
 
 
+def get_normalized_elevation_mask(world):
```
tcld: This doesn't return a mask.
Also, do you actually need to differentiate between land and ocean like this? If not, this method could be done in a single line. numpy.interp(world.elevation['data'], [world.elevation['data'].min(), world.elevation['data'].max()], [0, 255])

tcld: Converting to int and rounding would need to be done, too.

worldengine/draw.py
```diff
@@ -129,6 +210,109 @@ def elevation_color(elevation, sea_level=1.0):
...
+        The basic rules regarding noise generation are:
+        - Oceans have no noise added
+        - land tiles start with noise somewhere inside (-NOISE_RANGE, NOISE_RANGE) for each rgb value
+        - land tiles with high elevations further modify the noise by set amounts (to drain some of the 
+          color and make the map look more like mountains) 
+
+        The biome's base color may be interpolated with a predefined mountain brown color if the elevation is high enough.
+
+        Finally, the noise plus the biome color are added and returned
+    '''
```
tcld: Love the comments.

worldengine/draw.py
```diff
@@ -129,6 +210,109 @@ def elevation_color(elevation, sea_level=1.0):
...
+    # Default is no noise - will be overwritten if this tile is land
+    noise = (0, 0, 0)
+
+    if world.is_land((x, y)):
+        ## Generate some random noise to apply to this pixel
+        #  There is noise for each element of the rgb value
+        #  This noise will be further modified by the height of this tile
+        noise = (random.randint(-NOISE_RANGE, NOISE_RANGE), 
+                 random.randint(-NOISE_RANGE, NOISE_RANGE), 
+                 random.randint(-NOISE_RANGE, NOISE_RANGE))
```
tcld: Use numpy.randint instead. Same signature.

worldengine/draw.py
```diff
@@ -129,6 +210,109 @@ def elevation_color(elevation, sea_level=1.0):
...
+                    noise[1] + HIGH_MOUNTAIN_NOISE_MODIFIER[1], \
+                    noise[2] + HIGH_MOUNTAIN_NOISE_MODIFIER[2]
+
+            # Average the biome's color with the MOUNTAIN_COLOR to tint the terrain
+            biome_r = int((biome_r + MOUNTAIN_COLOR[0])/2)
+            biome_g = int((biome_g + MOUNTAIN_COLOR[1])/2)
+            biome_b = int((biome_b + MOUNTAIN_COLOR[2])/2)
+        #################################################
+
+        ####### Case 1 - elevation is high ########
```
tcld: This would be case 2.

worldengine/draw.py
```diff
@@ -129,6 +210,109 @@ def elevation_color(elevation, sea_level=1.0):
...
+
+        Finally, the noise plus the biome color are added and returned
+    '''
+    v = world.biome[y, x]
+    biome_r, biome_g, biome_b = _biome_satellite_colors[v]
+
+    # Default is no noise - will be overwritten if this tile is land
+    noise = (0, 0, 0)
+
+    if world.is_land((x, y)):
```
tcld: As suggested by @esampson, don't forget about the else case. In there you could check the temperature map and 'freeze' the ocean if necessary.

worldengine/draw.py
```diff
@@ -209,26 +393,93 @@ def draw_riversmap(world, target):
 
 
 def draw_grayscale_heightmap(world, target):
-    e = world.elevation['data']
+    c = get_normalized_elevation_mask(world)
```
tcld: Ah, now I understand why you mimicked this behavior. Well, I guess it's fine then.

worldengine/draw.py
```diff
@@ -209,26 +393,93 @@ def draw_riversmap(world, target):
...
+                        # Don't include ocean in the smoothing, if this tile happens to border an ocean
+                        if world.is_land((i, j)):
+                            # Grab each rgb value and append to the list
+                            r, g, b, a = target.pixels[i, j]
+                            all_r.append(r)
+                            all_g.append(g)
+                            all_b.append(b)
+
+                # Making sure there is at least one valid tile to be smoothed before we attempt to average the values
+                if all_r:
```
tcld: Would you mind doing the full check like 'len(...) > 0'?
I have seen some unexpected behavior due to this kind of abbreviated check.

worldengine/draw.py
```diff
@@ -209,26 +393,93 @@ def draw_riversmap(world, target):
...
+
+    # "Shade" the map by sending beams of light west to east, and increasing or decreasing value of pixel based on elevation difference
+    for y in range(SAT_SHADOW_SIZE-1, world.height-SAT_SHADOW_SIZE-1):
+        for x in range(SAT_SHADOW_SIZE-1, world.width-SAT_SHADOW_SIZE-1):
+            if world.is_land((x, y)):
+                r, g, b, a = target.pixels[x, y]
+                
+                # Build up list of elevations in the previous n tiles, where n is the shadow size.
+                # This goes left to right, so it would be the previous tiles on the same y level  
+                prev_elevs = [ world.elevation['data'][y-n, x-n] for n in range(1, SAT_SHADOW_SIZE+1) ]
```
tcld: You should probably use slice notation here, it's faster and will return a numpy array, should you need it.

tcld: Maybe I am misunderstanding this line of code at the moment.

tcld:
Aside from small comments this looks really, really good. Should you need help ironing out my 'complaints', just ask. :)

worldengine/generation.py
psi29a: `numpy.uint32.max` <-- usable?

tcld: That doesn't work for me. I also wasn't able to find wherever it said that randint() takes uint32 at most. (I think it was an error-message I got at some point after using a number that was too large.)
I'm not sure what to do about this.

psi29a:
```
In [24]: numpy.iinfo(numpy.uint32).max
Out[24]: 4294967295
```

tcld: Oh, that seems nice. I guess we should use ```numpy.iinfo(numpy.int32).max``` then?

psi29a: I think uint32 is fine, numpy doesn't seem to have a problem.

```
In [15]: numpy.random.randint(0,numpy.iinfo(numpy.uint32).max)
Out[15]: 250145318
```

Numpy only cries when using uint64:
```
In [14]: numpy.random.randint(0,numpy.iinfo(numpy.uint64).max)
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
<ipython-input-14-311a6f00f03d> in <module>()
----> 1 numpy.random.randint(0,numpy.iinfo(numpy.uint64).max)

/usr/lib/python2.7/dist-packages/numpy/random/mtrand.so in mtrand.RandomState.randint (numpy/random/mtrand/mtrand.c:7601)()

OverflowError: Python int too large to convert to C long
```

tcld: But @pangal said that he had some kind of problem. Are you using a 32 bit-installation of Python?

psi29a: @pangal what was the issue exactly? Something you can try to duplicate and paste here?

worldengine/draw.py
tcld: Since now I have a bit more time to explain: random.randint() gives different results depending on whether you use Python 2 or 3, which will result in (slightly) different worlds for both versions. We switched to numpy.random due to this.

worldengine/draw.py
```diff
@@ -129,6 +210,109 @@ def elevation_color(elevation, sea_level=1.0):
...
+        #################################################
+
+        ####### Case 3 - elevation is somewhat high ########
+        elif elev > HIGH_HILL_ELEV:   
+            # Make the random noise somewhat darker, and drain a little bit of green
+            noise = noise[0] - HIGH_HILL_NOISE_MODIFIER[0], \
+                    noise[1] - HIGH_HILL_NOISE_MODIFIER[1], \
+                    noise[2] - HIGH_HILL_NOISE_MODIFIER[2]
+
+        ####### Case 3 - elevation is a little bit high ########
```
tcld: And case 4.

worldengine/cli/main.py
psi29a: I'm OK with it. It does look gorgeous. :)

worldengine/draw.py
psi29a: Agreed!  :D

worldengine/draw.py
tcld: To follow up on my own idiocy: Maybe freezing oceans would be handled by its own function? In my head it seems fairly simple, but I wouldn't know where this would go in your code.
```
if temperature_at(x,y) < freezing and world.is_ocean(x,y):  # freezing refers to whatever temperature the cool biomes have (there is snow there, after all); probably has to be determined first
    r += (255 - r) / 2  # just brightening up the color by a lot
    g += (255 - r) / 2
    b += (255 - r) / 2

    temp_delta = freezing - temperature_at(x,y)
    if temp_delta < 0.02:  # some minimum temperature difference; has to be determined first
        if numpy.rand() > 0.6:  # if too close to the freezing threshold, only random portions (here 40%) of water will freeze
            map[x, y] = (r,g,b)
    else:
        map[x, y] = (r,g,b)
```
Just an example (where I might have overlooked something). The code could be simplified a bit.

ftomassetti:
I think freezing oceans would be a very nice addition but perhaps we could that in a second step. It would be cool to address some of the issues, merge the patch and then have a few incremental improvements.
In this way we could have this very cool feature going in the next release and then improve it later.

@pangal would be ok for you if we list you among the contributors?

worldengine/draw.py
rbb: Just a thought: Here on earth, oceans freeze at a lower temperature, primarily because of the salt in the ocean, and waves. I could believe that we'd want to mimic this behavior, with a separate threshold, or even use some % of the existing biome threshold. 

I wondered if things could ever go the other way. Doing some quick net searching, it looks like the freezing point can usually only be lowered by adding impurities (salt, alcohol, etc). Heavy water (deuterium oxide) has a higher freezing point, but doesn't change it much (3.82 °C).

worldengine/generation.py
pangal: If I change it to the number it was previously (Windows 10 64-bit, 32-bit python install) I get `Overflow error: Python int too large to convert to C long`

worldengine/draw.py
pangal: Good catch, will change!

worldengine/generation.py
tcld: So could you put this in instead of the fixed number?
```numpy.iinfo(numpy.int32).max```

File: worldengine/draw.py (hunk 13 above)
Comment (pangal) [inline worldengine/draw.py]: I tried sending the shadow west -> east, but I felt like the results looked better northwest -> southeast - so yes, moving diagonally is intentional, but it's up for debate.

File: worldengine/generation.py (hunk 1 above)
Comment (psi29a) [inline worldengine/generation.py]: I got that error message too, but only for uint64. This is only a problem for numpy, not python itself. 

I find it odd that people install 32-bit python on 64-bit windows. If you had used 64-bit python, this wouldn't be a problem. In addition to this, when we go to make official releases, it will be 64-bits only.

Comment (tcld) [inline worldengine/generation.py]: @pangal Which version of numpy are you using?

Comment (tcld) [inline worldengine/generation.py]: @psi29a Shouldn't we still just use int32? It would be a negligible loss for a medium-sized gain (less different seeds <-> reproducible worlds for everyone; especially the developers probably won't use a release version). Since you found a way to properly pull the correct number from somewhere, all would be fine in my opinion.

Comment (pangal) [inline worldengine/generation.py]: I'm using 32-bit python because some other libraries I'm using (libtcod, I think, and perhaps pygame) don't support 64-bit python. 

I can't check right now, but I'm using whatever the latest windows binary numpy has on sourceforge. 

I could revert this change, and then keep it changed locally whenever I want to use this library.

Comment (tcld) [inline worldengine/generation.py]: I think we should find a solution that just works for everybody. Using a different range will result in a different RNG-output and thus in different worlds (same geometry, different everything else).

**Comment by pangal** [comment]
I'd be honored to be listed as a contributor!

I may see if I get reasonable results by freezing the ocean when I have time later today.

File: worldengine/generation.py (hunk 1 above)
Comment (psi29a) [inline worldengine/generation.py]: :( I don't like the idea that we have to limit the random int pool size because another lib that we don't use.

Comment (tcld) [inline worldengine/generation.py]: Well, we could choose uint16 to be on par with the platec seed-range.^^

Comment (psi29a) [inline worldengine/generation.py]: *falls over*

File: worldengine/draw.py (hunk 5 above)
Comment (pangal) [inline worldengine/draw.py]: I understand the concept, but I don't really understand how to use the function. Is it documented anywhere? If this function exists, why are you suggesting to use the code you wrote out? What does the `[3]` in that code mean?

Comment (tcld) [inline worldengine/draw.py]: Take a look at world.py for an explanation of the 3, I think that's where that list is created. It is basically a list of percentages that are you to calculate mountain-heights, hill-heights etc.

In your case, since you are working with normalized values, you could bypass calling find_threshold() and instead manually calculate the necessary value with the line of code I gave above.

Comment (tcld) [inline worldengine/draw.py]: Highest ten percent of the map? 0.9*255, and so on.

Comment (tcld) [inline worldengine/draw.py]: The function I mentioned can be found [here](https://github.com/Mindwerks/worldengine/blob/master/worldengine/simulations/basic.py). You ask it *"What is the absolute value in height when I only want to look at the highest 10% of the map?"* and it'll answer *"229."* in case you provided it with an elevation-map normalized to the range [0,255].
This can be done with any kind of map (on the temperature-map you would look for the hottest spots, etc.). Normally the function is used because the exact range of values available for a single world is not exactly known. Elevations might range from [0,10.5] or maybe [0.1, 16].

I hope that wasn't too convoluted.

**Comment by psi29a** [comment]
Wooo... frozen oceans!

File: worldengine/draw.py [hunk 15]
```diff
@@ -203,32 +384,117 @@ def draw_riversmap(world, target):
 
     for y in range(world.height):
         for x in range(world.width):
-            target.set_pixel(x, y, sea_color if world.is_ocean((x, y)) else land_color)
+            target.set_pixel(x, y, sea_color if world.is_ocean((x, y)) else land_color)
```
tcld: These spaces look a bit unnecessary.^^

tcld:
Looks good now. The only things I noticed were a couple unnecessary spaces that were added and the missing newline at the end of draw.py.

ftomassetti:
Travis took over 1h to run so it was killed. Restarting it

ftomassetti:
I mean, I would restart it if it would let me do so... However I do not see any reasons why the build should be broken on Windows by this.

worldengine/cli/main.py
ftomassetti: I would also be fine with that

ftomassetti:
I would like to thank you a lot @pangal for proposing this patch and for working on it, adding comments and following the suggestions. I would like to thank you also @tcld who carefully reviewed, commented and helped us having this patch in better shape. Time to merge, add @pangal to the contributors. And we should also add @tcld : plase @tcld add yourself to the README!

tcld:
I started a commit for the README but didn't know what to write yet. I will do it before the next release.

PR #182, opened 2015-11-04
Changed seed-range for platec-seeds to use numpy constant.

tcld:
Just a small change after @psi29a found out how to pull the correct values from numpy.


psi29a:
^_^;; all my fault!

PR #183, opened 2015-11-05
Changed maximum seed-value in generation.py

tcld:
Since the change was removed from #181 , I decided to put it in so it isn't forgotten about.

### DIFF
--- a/worldengine/generation.py
+++ b/worldengine/generation.py
@@ -200,7 +200,7 @@ def generate_world(w, step):
 
     # Prepare sufficient seeds for the different steps of the generation
     rng = numpy.random.RandomState(w.seed)  # create a fresh RNG in case the global RNG is compromised (i.e. has been queried an indefinite amount of times before generate_world() was called)
-    sub_seeds = rng.randint(0, 4294967295, size=100)  # sys.maxsize didn't quite work
+    sub_seeds = rng.randint(0, numpy.iinfo(numpy.int32).max, size=100)  # choose lowest common denominator (32 bit Windows numpy cannot handle a larger value)
     seed_dict = {
                  'PrecipitationSimulation': sub_seeds[ 0],  # after 0.19.0 do not ever switch out the seeds here to maximize seed-compatibility
                  'ErosionSimulation':       sub_seeds[ 1],


worldengine/generation.py
```diff
@@ -200,7 +200,7 @@ def generate_world(w, step):
 
     # Prepare sufficient seeds for the different steps of the generation
     rng = numpy.random.RandomState(w.seed)  # create a fresh RNG in case the global RNG is compromised (i.e. has been queried an indefinite amount of times before generate_world() was called)
-    sub_seeds = rng.randint(0, 4294967295, size=100)  # sys.maxsize didn't quite work
+    sub_seeds = rng.randint(0, numpy.iinfo(numpy.int32).max, size=100)  # choose lowest common denominator (32 bit Python seemingly cannot handle a larger value)
```
psi29a: 32-bit Python can handle it just fine, it is 32-bit windows Numpy that has issues.
So we go from 4294967295 to 2147483647 which is a real pity. :(

tcld: We only have 65536 differently shaped worlds at the moment with 2147483647 different temperature-, precipitation- and permeability-maps on top of it (the other simulations don't make use of the RNG). I would rather feel sad about the uint16-limitation of platec.^^

(I'll change the coment.)

psi29a: I'm not arguing! :D 

I just hate numpy's support of 32-bit windows.

psi29a: numpy.random.randint(0,numpy.iinfo(numpy.int32).max-1)
^-- this should work

psi29a: @pangal can you test that the above works for you in windows with 32-bit python/numpy?

tcld: For me this happens:
```python
Python 3.4.3 (default, Oct 14 2015, 20:28:29) 
[GCC 4.8.4] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import numpy
>>> numpy.iinfo(numpy.int32).max
2147483647
```
That is an odd number, so it has to be 2^31-1 already. What would subtracting ```1``` do?

psi29a: I meant uint32... but testing it doesn't get around the issue. int32 is good enough for now. ;)

tcld:
By the way: I did use the same value that @pangal proposed here: https://github.com/pangal/worldengine/commit/9a35e5d2a63da0ddfe4ee5f93ae1c37bbb9b2d21
So I would assume that value actually worked for him and we don't have to wait for him to test it again.

PR #185, opened 2015-11-05
New Tests; Several Bug-Fixes

tcld:
I tried to add a test for the satellite-view (and added a couple lines of code to properly test the scatter-plot).

I think I might have chosen a bad time for this since @pangal might want to continue work on this himself...but I thought this might be a quick change. It isn't, sadly, so far I haven't gotten the code to pass the satellite-test. I don't really know why, but maybe somebody sees the flaw.

EDIT: Here the data-PR https://github.com/Mindwerks/worldengine-data/pull/12, even though this should **not be merged right now** due the failed final test.

EDIT2: If @pangal (or somebody else) wants to continue work on this, I could probably rebase this PR, so please don't hesitate. :)

**EDIT3:** This is now based on #163.

### DIFF
--- a/tests/draw_test.py
+++ b/tests/draw_test.py
@@ -3,7 +3,7 @@ import os
 import numpy
 from worldengine.draw import _biome_colors, draw_simple_elevation, elevation_color, \
     draw_elevation, draw_riversmap, draw_grayscale_heightmap, draw_ocean, draw_precipitation, \
-    draw_world, draw_temperature_levels, draw_biome, draw_scatter_plot
+    draw_world, draw_temperature_levels, draw_biome, draw_scatter_plot, draw_satellite
 from worldengine.biome import Biome
 from worldengine.world import World
 from worldengine.image_io import PNGWriter, PNGReader
@@ -150,8 +150,15 @@ class TestDraw(TestBase):
 
     def test_draw_scatter_plot(self):
         w = World.open_protobuf("%s/seed_28070.world" % self.tests_data_dir)
-        target = PNGWriter.rgba_from_dimensions(16, 16)
-        draw_scatter_plot(w, 16, target)
+        target = PNGWriter.rgba_from_dimensions(512, 512)
+        draw_scatter_plot(w, 512, target)
+        self._assert_img_equal("scatter_28070", target)
+
+    def test_draw_satellite(self):
+        w = World.open_protobuf("%s/seed_28070.world" % self.tests_data_dir)
+        target = PNGWriter.rgba_from_dimensions(w.width, w.height)
+        draw_satellite(w, target)
+        self._assert_img_equal("satellite_28070", target)
 
 if __name__ == '__main__':
     unittest.main()
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -217,7 +217,11 @@ def elevation_color(elevation, sea_level=1.0):
 def add_colors(*args):
     ''' Do some *args magic to return a tuple, which has the sums of all tuples in *args '''
     # Adapted from an answer here: http://stackoverflow.com/questions/14180866/sum-each-value-in-a-list-of-tuples
-    return tuple([sum(x) for x in zip(*args)])
+    added = [sum(x) for x in zip(*args)]
+
+    for i in range(len(added)):  # prevent over- and underflows
+        added[i] = max(min(added[i], 255), 0)  # restrict to uint8
+    return added
 
 def average_colors(c1, c2):
     ''' Average the values of two colors together '''
@@ -307,9 +311,10 @@ def get_biome_color_based_on_elevation(world, elev, x, y, rng):
     modification_amount = int(elev / BASE_ELEVATION_INTENSITY_MODIFIER)
     base_elevation_modifier = (modification_amount, modification_amount, modification_amount)
 
-    this_tile_color = add_colors(biome_color, noise, base_elevation_modifier)  # this can lead to an overflow, i.e. a value above 255
-
+    this_tile_color = add_colors(biome_color, noise, base_elevation_modifier)
     return this_tile_color
+
+
 # ----------------------
 # Draw on generic target
 # ----------------------
@@ -367,7 +372,7 @@ def draw_grayscale_heightmap(world, target):
 
 
 def draw_satellite(world, target):
-    ''' This draws a "satellit map" - a view of the generated planet as it may look from space '''
+    ''' This draws a "satellite map" - a view of the generated planet as it may look from space '''
 
     # Get an elevation mask where heights are normalized between 0 and 255
     elevation_mask = get_normalized_elevation_array(world)
@@ -388,7 +393,6 @@ def draw_satellite(world, target):
             # the map is smoothed and shaded.
             target.set_pixel(x, y, (r, g, b, 255))
 
-
     # Loop through and average a pixel with its neighbors to smooth transitions between biomes
     for y in range(1, world.height-1):
         for x in range(1, world.width-1):
@@ -405,7 +409,7 @@ def draw_satellite(world, target):
                         # Don't include ocean in the smoothing, if this tile happens to border an ocean
                         if world.is_land((i, j)):
                             # Grab each rgb value and append to the list
-                            r, g, b, a = target.pixels[i, j]
+                            r, g, b, a = target[i, j]
                             all_r.append(r)
                             all_g.append(g)
                             all_b.append(b)
@@ -419,30 +423,28 @@ def draw_satellite(world, target):
                     ## Setting color of the pixel again - this will be once more modified by the shading algorithm
                     target.set_pixel(x, y, (avg_r, avg_g, avg_b, 255))
 
-
     ## After smoothing, draw rivers
     for y in range(world.height):
         for x in range(world.width):
             ## Color rivers
             if world.is_land((x, y)) and (world.river_map[x, y] > 0.0):
-                base_color = target.pixels[x, y]
+                base_color = target[x, y]
 
                 r, g, b = add_colors(base_color, RIVER_COLOR_CHANGE)
                 target.set_pixel(x, y, (r, g, b, 255))
 
             ## Color lakes
             if world.is_land((x, y)) and (world.lake_map[x, y] != 0):
-                base_color = target.pixels[x, y]
+                base_color = target[x, y]
 
                 r, g, b = add_colors(base_color, LAKE_COLOR_CHANGE)
                 target.set_pixel(x, y, (r, g, b, 255))
-
 
     # "Shade" the map by sending beams of light west to east, and increasing or decreasing value of pixel based on elevation difference
     for y in range(SAT_SHADOW_SIZE-1, world.height-SAT_SHADOW_SIZE-1):
         for x in range(SAT_SHADOW_SIZE-1, world.width-SAT_SHADOW_SIZE-1):
             if world.is_land((x, y)):
-                r, g, b, a = target.pixels[x, y]
+                r, g, b, a = target[x, y]
                 
                 # Build up list of elevations in the previous n tiles, where n is the shadow size.
                 # This goes northwest to southeast
@@ -460,9 +462,9 @@ def draw_satellite(world, target):
                 # The amplified difference is now translated into the rgb of the tile.
                 # This adds light to tiles higher that the previous average, and shadow
                 # to tiles lower than the previous average
-                r += adjusted_difference
-                g += adjusted_difference
-                b += adjusted_difference
+                r = max(min(adjusted_difference + r, 255), 0)  # NOTE: This could cause underflows and overflows
+                g = max(min(adjusted_difference + g, 255), 0)
+                b = max(min(adjusted_difference + b, 255), 0)
 
                 # Set the final color for this pixel
                 target.set_pixel(x, y, (r, g, b, 255))
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -409,7 +409,7 @@ def draw_satellite(world, target):
                         # Don't include ocean in the smoothing, if this tile happens to border an ocean
                         if world.is_land((i, j)):
                             # Grab each rgb value and append to the list
-                            r, g, b, a = target[i, j]
+                            r, g, b, a = target[j, i]
                             all_r.append(r)
                             all_g.append(g)
                             all_b.append(b)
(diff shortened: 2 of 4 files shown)


tests/draw_test.py
```diff
@@ -164,8 +164,15 @@ def test_draw_biome(self):
 
     def test_draw_scatter_plot(self):
         w = World.open_protobuf("%s/seed_28070.world" % self.tests_data_dir)
-        target = PixelCollector(16, 16)
-        draw_scatter_plot(w, 16, target)
+        target = PixelCollector(512, 512)
+        draw_scatter_plot(w, 512, target)
```
tcld: The generated PNG isn't exactly large so I guess the size is ok.

worldengine/draw.py
```diff
@@ -307,7 +307,7 @@ def get_biome_color_based_on_elevation(world, elev, x, y):
     modification_amount = int(elev / BASE_ELEVATION_INTENSITY_MODIFIER)
     base_elevation_modifier = (modification_amount, modification_amount, modification_amount)
 
-    this_tile_color = add_colors(biome_color, noise, base_elevation_modifier)
+    this_tile_color = add_colors(biome_color, noise, base_elevation_modifier)  # this can lead to an overflow, i.e. a value above 255
```
tcld: I think this might not be perfect yet. Maybe an overflow wouldn't even cause problems, I am not sure.

worldengine/draw.py
```diff
@@ -403,6 +403,8 @@ def draw_satellite(world, target):
     # Get an elevation mask where heights are normalized between 0 and 255
     elevation_mask = get_normalized_elevation_array(world)
 
+    rng = numpy.random.RandomState(world.seed)  # create our own random generator; necessary for now to make the tests reproducible, even though it is a bit ugly
```
tcld: Since the test-suite can call this function, I had to do something about the RNG to give the test a chance to pass. The remaining problem doesn't seem to be random.

tcld:
Here the test that fails for me (using the updated data):
```
self = <tests.draw_test.TestDraw testMethod=test_draw_satellite>

    def test_draw_satellite(self):
        w = World.open_protobuf("%s/seed_28070.world" % self.tests_data_dir)
        target = PixelCollector(w.width, w.height)
        draw_satellite(w, target)
>       self._assert_img_equal("satellite_28070", target)

tests/draw_test.py:175: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
tests/draw_test.py:70: in _assert_img_equal
    % (x, y, blessed_pixel, drawn_pixel))
E   AssertionError: Tuples differ: (244, 244, 251, 255) != (250, 255, 265, 255)
E   
E   First differing element 0:
E   244
E   250
E   
E   - (244, 244, 251, 255)
E   + (250, 255, 265, 255) : Pixels at 39, 16 are different. Blessed (244, 244, 251, 255), drawn (250, 255, 265, 255)
```

No idea yet why this happens, but it is reproducible (I tried three times, the results were exactly the same every time).

ftomassetti:
I am not sure about this error, I guess it could be due to extraction of random values but that is it

tcld:
I think I traced down the problem:
```python
r, g, b, a = target.pixels[x, y]  # target is a ImagePixelSetter
r, g, b = test[y, x]  # test is a numpy-array with no special data type (I assume int32)
```
Both "arrays" have been written using exactly the same code, their content at several pixels differs, though.

Error-message of the test (the first values stem from a proper PNG-file, the latter from the actively drawn test-image):
```Blessed (244, 244, 251, 255), drawn (250, 255, 265, 255)  # ImagePixelSetter
Blessed (250, 255, 255, 255), drawn (250, 255, 265, 255)  # numpy-aray```

So two issues crop up: An overflow is happening (3rd value) and the read-out of the ImagePixelSetter returns a different value than the numpy-array.

**Comment by tcld** [comment]
Most of the problems came from over-/underflowing uint8s that were written to the PNG. But even after fixing those I am left with a failed test due to differing pixels (not caused by the RNG, not caused by rounding and not caused by over-/underflows).

I tried rebasing this PR to #163 while searching for the cause of the problem and that worked fine. Now I am a little bit unsure if it is worth the time trying to find the problem in a piece of code that I kind of wanted to see replaced in the first place. Would it be ok if I just put this on top of #163 (with some small adjustments)? (That question is especially directed at @psi29a.)

My latest findings are in here, although the (new) tests still cannot be passed.

**Comment by psi29a** [comment]
Put it on top of #163 , we we're ready I'll merge it in.  Having pypng as an additional depend isn't a problem as it is pure python.

**Comment by tcld** [comment]
Ok.

**Comment by tcld** [comment]
I had to take care of some other things first. Rebase is now done, all tests are passed for me (Python 3).

@esampson: Would you mind taking a look at the scatter-plot I generated [here](https://github.com/tcld/worldengine-data/blob/tests_upd/tests/images/scatter_28070.png)? It looks quite different from the old one. It's not due to changes in the code (I checked using the old world, the output was the same); hopefully this is fine.

**Another remark:**
The [precipitation](https://github.com/tcld/worldengine-data/blob/tests_upd/tests/images/precipitation_28070.png)-map now looks a lot better, the [temperature](https://github.com/tcld/worldengine-data/blob/tests_upd/tests/images/temperature_28070.png)-map is slightly less homogeneous. I don't really know where these differences come from compared to the old map, it might be due to some rounding that is performed. I only wanted to mention it since at some point @psi29a noticed that the those maps were a little too homogeneous and that this had been different in the past.

**Comment by ftomassetti** [comment]
Stupid question: why we inverted in almost all the arrays x and y? I agree we should be consisteng but wouldn't make sense to keep the order x, y?

**Comment by tcld** [comment]
There was a discussion about that, I think it is visible to you right in this PR (above your post). If you can't see it, I could summarize it.

**Comment by ftomassetti** [comment]
I realize you spent a lot of time inverting x, y to y, x and I really agree we should stick to one order in all the places, I just think as @pangal that x, y seems more natural than y, x. @psi29a any opinion on this?

**Comment by tcld** [comment]
Well, how about going for [y, x] now (at least it would be consistent) and then fully switching to [x, y] in a later PR.

The old arrays were all initialized with ```array = [[0 for x in range(width)] for y in range(height)]``` and then accessed via ```array[y][x]```. When I switched to numpy I had to adhere to that or things would have probably exploded.
The initial reasoning behind ```[y][x]``` might just be that ```[[0 for x in range(width)] for y in range(height)]``` is a very natural (i.e. western) way of looking at matrices: Rows are appended to form a full matrix (read from left-to-right, then top-to-bottom, as with books); the exact same way numpy does it (as the code-snippets above show).

**Comment by ftomassetti** [comment]
I would normally be for splitting things among different PRs but changing
all the array access and to change them immediately after seems to be way
too complicate. I would say, let's just pick on way of access the arrays
and stick with it. My preference is for [x, y] because it is more natural.
Is there any strong reason to go for the opposite? Does it offer way better
performances? Is the problem that [y, x] is more common?

I understand that when using separate indexes it makes sense to use [y][x]
instead of [x][y]. Is in numpy arrays [x, y] just a shorctu for [x][y]?

On Mon, Nov 9, 2015 at 3:32 PM, tcld <notifications@github.com> wrote:

> Well, how about going for [y, x] now (at least it would be consistent) and
> then fully switching to [x, y] in a later PR.
>
> The old arrays were all initialized with array = [[0 for x in
> range(width)] for y in range(height)] and then accessed via array[y][x].
> When I switched to numpy I had to adhere to that or things would have
> probably exploded.
> The initial reasoning behind [y][x] might just be that [[0 for x in
> range(width)] for y in range(height)] is a very natural (i.e. western)
> way of looking at matrices: Rows are appended to form a full matrix (read
> from left-to-right, then top-to-bottom, as with books); the exact same way
> numpy does it (as the code-snippets above show).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Mindwerks/worldengine/pull/185#issuecomment-155079175>
> .
>



-- 
Website at http://tomassetti.me <http://www.federico-tomassetti.it>
GitHub https://github.com/ftomassetti

**Comment by tcld** [comment]
The numpy-notation [y, x] is equivalent to [y][x].

Is there a strong reason to go with [x, y]? I found it useful to use [y, x] when printing a numpy-array for debug-purposes. x in that case refers to the "natural" x-coordinate. Other than that it probably doesn't really matter. Although a full switch will cause saves to become incompatible again, so it should be done sooner rather than later.

I think the benefits of this kind of cosmetic change are not worth the work. There are at least a couple lines of code in almost every source-file, very many lines in some, that would need to be changed. There is no speed to be gained and no significant simplification of code.

**Comment by rbb** [comment]
If you want to print a numpy array, with swapped rows and colums (in iPython for example), you can always use numpy.transpose(). 

For example: 
```python
myVar = numpy.rand(x,y)
print( myVar.transpose() )
```

ftomassetti:
ok, you have convinced me, I am fine with [y, x] unless someone else has strong opinions about that

pangal:
I strongly believe it should be [x, y] to keep convention with standard Cartesian coords - but I understand that it's a big undertaking to rewrite it, so I'm ok with [y, x] if it's consistent throughout the codebase.

psi29a:
Agreed with @pangal pangal over being consistent.

We are mixing Image nomenclature with Matrix nomenclature, Numpy uses the Matrix (I,J) nomenclature.
```
 Z = np.zeros((2,4))
print(Z)
[[ 0.  0.  0.  0.]
 [ 0.  0.  0.  0.]]
```

while Images (on screen) use the X,Y (top left to bottom right).

We shouldn't be confusing it more with Cartesian since we are not using graphs. ;)

As @tcld said, when dealing with Images, we should take the Numpy array and transpose it. However all work should remain in Numpy native (I,J) matrix format.

tcld:
Ah, I was looking for something like this but didn't find the right keyword.

psi29a:
I meant, that when we go about transforming our matrix data into image data to be written out to disk we can transpose the numpy matrix as @rbb mentioned above. Or do you consider this unnecessary?

If we can hammer this out, we can get WE into a more stable/common practice state. 

This would be something to mention in our documentation as well and in comments for people who come after us. :)

tcld:
So the only place where image-notation should even be considered is *inside* the class that handles writing/reading images? I just wanted to be very clear about it. (Transposing right before writing isn't necessary, by the way, since the arrays need to be flattened first anyway. Same goes for reading.)

I could add some comments about this to the image-handler classes, but since so many places are affected by this choice, a tiny section in the documentation is probably important.

psi29a:
You're correct and right! Comments in code is the way to go. :)

psi29a:
This needs to have the manifest updated and images to test...

tcld:
The images have been [ready for a while](https://github.com/Mindwerks/worldengine-data/pull/12). Updated manifest? Why that?

psi29a:
Without an updated manifest, we can't make a release.
https://travis-ci.org/Mindwerks/worldengine/jobs/90341676

^-- points out which files are missing. If it has nothing to do with python or the library's functionality, I usually just ignore it.

tcld:
Oh. I mentioned that before:
https://github.com/Mindwerks/worldengine/pull/179

I don't know how to properly fix it, and it's not related to this PR.

psi29a:
Right. I'll take care of it when I get the chance.

toddcarnes:
OK. I have to ask. What does PR stand for in this context?

tcld:
Whenever I use it, it stands for *Pull Request*. When others use it, I have always assumed the same and that seemed to work so far.

toddcarnes:
Thank you

psi29a:
Some people say Pull Request (PR) and others Merge Request (MR). Github has the the workflow that it will ask to merge your pull request if you have the right privileges.

tcld:
I rebased this and now the changes are kind of trivial. :)

tcld:
How about this one, @psi29a and/or @ftomassetti ? It's only about minor things, but it has been done for a while.
Data: https://github.com/Mindwerks/worldengine-data/pull/12

tcld:
The changes to lines 400 to 450 in `draw.py` are actually important; I am not sure why they weren't part of the PR that previously unified the matrix-accessors to [y, x]. Maybe I overlooked it after rebasing onto the satellite-commit.

worldengine/draw.py
```diff
@@ -367,11 +372,13 @@ def draw_grayscale_heightmap(world, target):
 
 
 def draw_satellite(world, target):
-    ''' This draws a "satellit map" - a view of the generated planet as it may look from space '''
+    ''' This draws a "satellite map" - a view of the generated planet as it may look from space '''
```
ftomassetti: Perhaps we could specify what type of object should target be?

worldengine/draw.py
```diff
@@ -458,9 +462,9 @@ def draw_satellite(world, target):
                 # The amplified difference is now translated into the rgb of the tile.
                 # This adds light to tiles higher that the previous average, and shadow
                 # to tiles lower than the previous average
-                r += adjusted_difference
-                g += adjusted_difference
-                b += adjusted_difference
+                r = max(min(adjusted_difference + r, 255), 0)  # NOTE: This could cause underflows and overflows
```
ftomassetti: do we want to create an helper function for checking the value is in range?

ftomassetti:
A few comments added but in general it looks good

worldengine/draw.py
tcld: That would have be done for all drawing-functions - so maybe up at the top of the file?
Within this function:
`target` has to have a method `set_pixel()` and the pixels need to be accessible via `[y, x]`.

worldengine/draw.py
tcld: I think that function might end up as never being used by anybody else because it is so small. Maybe we can make use of [a numpy-function](http://docs.scipy.org/doc/numpy/reference/generated/numpy.clip.html) or something similar?

tcld:
I took care of one of the comments but didn't know what to do exactly about describing `target`.

ftomassetti:
I am running tests locally together with new data

ftomassetti:
Tests passed locally. Restarting them on Travis.

Issue #65, opened 2015-05-04
Make the protobuf format the default one


ftomassetti:
When generating a world by default it is saved in pickle format. The user has the possibility to save it in protobuf format if he chooses an option. I think it should be the other way around or we could also drop the pickle format and support exclusively the protobuf format.

psi29a:
Problem with python3, so I wouldn't make this the default.

ftomassetti:
Mmm, in the other end pickle is limiting (not much support for other languages, not particularly efficient). Should we think about other formats like HDF5?

psi29a:
For my world data, I used PyTables which used HDF5 to store the data. PyTables gives you a nice abstraction to work with. Bonus for being actively developed, widely used and supports Python2 and Python3.  I know it works on windows because I packaged WorldSynth with it.

https://pypi.python.org/pypi/tables

psi29a:
"PyTables is a package for managing hierarchical datasets and designed to efficently cope with extremely large amounts of data. PyTables is built on top of the HDF5 library and the NumPy package and features an object-oriented interface that, combined with C-code generated from Cython sources, makes of it a fast, yet extremely easy to use tool for interactively save and retrieve large amounts of data."

psi29a:
This also makes it easy to hand world files off to other developers because they can read it in via PyTables or even HDF5 to just have access to the data-sets themselves.

ftomassetti:
Now the protobuf is working of Python 3 should we make it the default format? I am also in favor of adopring HDF5, it just would require some more work while we can make protobuf the default very easily :)

ftomassetti:
Protobuf do not seem loved enough to be the default format, closing

psi29a:
who said this? Protobuf is loved enough to be the default format from my POV.

ftomassetti:
Sorry, my misunderstaning. However it would probably make sense to wait that Protobuf 3 is released for all the platforms.

psi29a:
Isn't Protobuf3 what is being pushed in a PR by @tcld ?

psi29a:
It's already here:
https://github.com/Mindwerks/worldengine/blob/master/requirements.txt#L6

:D

tcld:
It's in there because otherwise there wouldn't be any format that is equally supported by Python 2 and 3. And it works for now, but @ftomassetti said that there were some problems with Java (which requires a newer version?). Could it be that pip is just not using the latest version?

[...]

PR #187, opened 2015-11-06
pickle: protobuf: make protobuf the default format instead of pickle

ftomassetti:
Solve #65

### DIFF
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -248,11 +248,11 @@ def main():
              "a name is not provided, then seed_N.world, " +
              "where N=SEED",
         metavar="STR")
-    # TODO: add description of protocol buffer
-    parser.add_argument('-b', '--protocol-buffer', dest='protobuf',
+    # TODO: add description of pickle and protobuf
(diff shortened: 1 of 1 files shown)


psi29a:
Can we just drop 'official' support of pickle, as in in the params?  We leave the code to handle pickle worlds until 0.20 ?

ftomassetti:
sure, done

ftomassetti:
Cool, one to go before closing all the issues for v0.19.0

tcld:
@ftomassetti and @psi29a:
I just realized that removing the ability to save pickle-files kind of makes it impossible to generate test-worlds in the pickle-format.
So either there needs to be some kind of backdoor or the pickle-tests etc. should be removed, too. (I think only the latter makes sense since loading of old pickle-worlds will fail anyway once there has been any sort of change to the save files. So the end-user won't have any advantage from residual pickle-support).

psi29a:
The later is the best, it should also be purged. Tests can go away.

tcld:
I took some time and removed pickle completely: https://github.com/Mindwerks/worldengine/pull/192

PR #188, opened 2015-11-06
More Mountain Levels

tcld:
While I was at it I thought I might make the new satellite-map use the proper mountain-heights instead of the ones defined in draw.py - that way the output should coincide better with the other maps.

I then noticed that the mountain-thresholds were kind of improperly used. Hills were used as mountains and the higher mountainous areas probably never existed in any world because the values were too high.

Data for this PR can be found [here](https://github.com/Mindwerks/worldengine-data/pull/17).

### DIFF
--- a/worldengine/World.proto
+++ b/worldengine/World.proto
@@ -57,62 +57,65 @@ message World {
     required int32  height = 5;
 
     // Elevation
-    required DoubleMatrix heightMapData  = 6;
-    required double heightMapTh_sea      = 7;
-    required double heightMapTh_plain    = 8;
-    required double heightMapTh_hill     = 9;
+    required DoubleMatrix heightMapData       = 6;
+    required double heightMapTh_sea           = 7;
+    required double heightMapTh_plain         = 8;
+    required double heightMapTh_hill          = 9;
+    required double heightMapTh_low_mountain  = 10;
+    required double heightMapTh_med_mountain  = 11;
+    required double heightMapTh_high_mountain = 12;
 
     // Plates
-    required IntegerMatrix plates        = 10;
+    required IntegerMatrix plates          = 13;
 
     // Ocean
-    required BooleanMatrix ocean         = 11;
-    required DoubleMatrix  sea_depth     = 12;
+    required BooleanMatrix ocean           = 14;
+    required DoubleMatrix  sea_depth       = 15;
 
     // Biome
-    optional IntegerMatrix biome         = 13;
+    optional IntegerMatrix biome           = 16;
 
     // Humidity
-    optional DoubleMatrixWithQuantiles humidity = 14;
+    optional DoubleMatrixWithQuantiles humidity = 17;
 
     // Irrigation
-    optional DoubleMatrix irrigation  = 15;
+    optional DoubleMatrix irrigation       = 18;
 
     // Permeability
-    optional DoubleMatrix permeabilityData = 16;
-    optional double permeability_low       = 17;
-    optional double permeability_med       = 18;
+    optional DoubleMatrix permeabilityData = 19;
+    optional double permeability_low       = 20;
+    optional double permeability_med       = 21;
 
     // Watermap
-    optional DoubleMatrix watermapData     = 19;
-    optional double watermap_creek         = 20;
-    optional double watermap_river         = 21;
-    optional double watermap_mainriver     = 22;
+    optional DoubleMatrix watermapData     = 22;
+    optional double watermap_creek         = 23;
+    optional double watermap_river         = 24;
+    optional double watermap_mainriver     = 25;
 
     // Precipitation
-    optional DoubleMatrix precipitationData = 23;
-    optional double precipitation_low       = 24;
-    optional double precipitation_med       = 25;
+    optional DoubleMatrix precipitationData = 26;
+    optional double precipitation_low       = 27;
+    optional double precipitation_med       = 28;
 
     // Temperature
-    optional DoubleMatrix temperatureData    = 26;
-    optional double temperature_polar        = 27;
-    optional double temperature_alpine       = 28;
-    optional double temperature_boreal       = 29;
-    optional double temperature_cool         = 30;
-    optional double temperature_warm         = 31;
-    optional double temperature_subtropical  = 32;
+    optional DoubleMatrix temperatureData    = 29;
+    optional double temperature_polar        = 30;
+    optional double temperature_alpine       = 31;
+    optional double temperature_boreal       = 32;
+    optional double temperature_cool         = 33;
+    optional double temperature_warm         = 34;
+    optional double temperature_subtropical  = 35;
 
     // Data about generation:
     // introduced in v0.5.3
     // this is optional for backward compatibility reasons
-    optional GenerationData generationData = 33;
+    optional GenerationData generationData = 36;
 
-    optional DoubleMatrix lakemap = 34;
-    optional DoubleMatrix rivermap = 35;
+    optional DoubleMatrix lakemap = 37;
+    optional DoubleMatrix rivermap = 38;
 
     // Ice-caps
-    optional DoubleMatrix icecap  = 36;
+    optional DoubleMatrix icecap  = 39;
 }
 
 
--- a/worldengine/draw.py
+++ b/worldengine/draw.py
@@ -11,13 +11,6 @@ from worldengine.image_io import PNGWriter
 ### For draw_satellite ###
 NOISE_RANGE = 15 # a random value between -NOISE_RANGE and NOISE_RANGE will be added to the rgb of each pixel
 
-# These are arbitrarily-chosen elevation cutoffs for 4 different height levels. 
-# Some color modifiers will be applied at each level
-HIGH_MOUNTAIN_ELEV = 215
-MOUNTAIN_ELEV      = 175
-HIGH_HILL_ELEV     = 160
-HILL_ELEV          = 145
-
 # These are rgb color values which will be added to the noise, if the elevation is greater than the height specified
 # These are not cumulative
 HIGH_MOUNTAIN_NOISE_MODIFIER = (10, 6,   10)
@@ -279,30 +272,29 @@ def get_biome_color_based_on_elevation(world, elev, x, y, rng):
         ## Generate some random noise to apply to this pixel
         #  There is noise for each element of the rgb value
         #  This noise will be further modified by the height of this tile
-
         noise = rng.randint(-NOISE_RANGE, NOISE_RANGE, size=3)  # draw three random numbers at once
 
-        ####### Case 1 - elevation is very high ########
-        if elev > HIGH_MOUNTAIN_ELEV:     
-            # Modify the noise to make the area slightly brighter to simulate snow-topped mountains.
-            noise = add_colors(noise, HIGH_MOUNTAIN_NOISE_MODIFIER)
-            # Average the biome's color with the MOUNTAIN_COLOR to tint the terrain
-            biome_color = average_colors(biome_color, MOUNTAIN_COLOR)
+    ####### Case 1 - elevation is very high ########
+    if world.is_high_mountain((x, y)):
+        # Modify the noise to make the area slightly brighter to simulate snow-topped mountains.
+        noise = add_colors(noise, HIGH_MOUNTAIN_NOISE_MODIFIER)
+        # Average the biome's color with the MOUNTAIN_COLOR to tint the terrain
+        biome_color = average_colors(biome_color, MOUNTAIN_COLOR)
 
-        ####### Case 2 - elevation is high ########
-        elif elev > MOUNTAIN_ELEV:   
-            # Modify the noise to make this tile slightly darker, especially draining the green
-            noise = add_colors(noise, MOUNTAIN_NOISE_MODIFIER)
-            # Average the biome's color with the MOUNTAIN_COLOR to tint the terrain
-            biome_color = average_colors(biome_color, MOUNTAIN_COLOR)
+    ####### Case 2 - elevation is high ########
+    elif world.is_medium_mountain((x, y)):
+        # Modify the noise to make this tile slightly darker, especially draining the green
+        noise = add_colors(noise, MOUNTAIN_NOISE_MODIFIER)
+        # Average the biome's color with the MOUNTAIN_COLOR to tint the terrain
+        biome_color = average_colors(biome_color, MOUNTAIN_COLOR)
 
-        ####### Case 3 - elevation is somewhat high ########
-        elif elev > HIGH_HILL_ELEV:   
-            noise = add_colors(noise, HIGH_HILL_NOISE_MODIFIER)
+    ####### Case 3 - elevation is somewhat high ########
+    elif world.is_low_mountain((x, y)):
+        noise = add_colors(noise, HIGH_HILL_NOISE_MODIFIER)
 
-        ####### Case 4 - elevation is a little bit high ########
-        elif elev > HILL_ELEV:   
-            noise = add_colors(noise, HILL_NOISE_MODIFIER)
+    ####### Case 4 - elevation is a little bit high ########
+    elif world.is_hill((x, y)):
+        noise = add_colors(noise, HILL_NOISE_MODIFIER)
 
     # There is also a minor base modifier to the pixel's rgb value based on height
     modification_amount = int(elev / BASE_ELEVATION_INTENSITY_MODIFIER)
--- a/worldengine/drawing_functions.py
+++ b/worldengine/drawing_functions.py
@@ -100,10 +100,10 @@ def _find_mountains_mask(world, factor):
     _mask = numpy.zeros((factor * world.height, factor * world.width), dtype=float)
     for y in range(factor * world.height):
         for x in range(factor * world.width):
-            if world.is_mountain((int(x / factor), int(y / factor))):
+            if world.is_low_mountain((int(x / factor), int(y / factor)), include_higher=True):  # maybe use is_medium_mountain()?
                 v = len(world.tiles_around((int(x / factor), int(y / factor)),
                                            radius=3,
-                                           predicate=world.is_mountain))
+                                           predicate=lambda pos: world.is_low_mountain(pos, include_higher=True)))
                 if v > 32:
                     _mask[y, x] = v / 4.0  # force conversion to float, Python 2 will *not* do it automatically
     return _mask
--- a/worldengine/generation.py
+++ b/worldengine/generation.py
@@ -105,11 +105,14 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
     """
     e = world.elevation['data']
     ocean = fill_ocean(e, ocean_level)
-    pl = find_threshold_f(e, 0.70, ocean=ocean)  # the highest x% of land are declared plains
-    hl = find_threshold_f(e, 0.35, ocean=ocean)  # the highest x% are declared hills
-    ml = find_threshold_f(e, 0.10, ocean=ocean)  # the highest x% are declared low mountains
-    mml = find_threshold_f(e, 0.06, ocean=ocean)  # the highest x% are declared medium mountains
-    hml = find_threshold_f(e, 0.02, ocean=ocean)  # the highest x% are declared high mountains
+
+    # values chosen to emulate the exponential behavior seen here:
+    # http://www.ngdc.noaa.gov/mgg/global/etopo1_surface_histogram.html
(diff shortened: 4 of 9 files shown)


tcld:
This has been updated and now seems to work as intended. I will add some comments to the important changes, one of them needs some feedback and fine-tuning for sure.

worldengine/generation.py
```diff
@@ -124,12 +124,17 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
     """
     e = world.elevation['data']
     ocean = fill_ocean(e, ocean_level)
-    hl = find_threshold_f(e, 0.10)
-    ml = find_threshold_f(e, 0.03)
+    pl = find_threshold_f(e, 0.70, ocean=ocean)  # the highest 70% of land are declared plains
+    hl = find_threshold_f(e, 0.35, ocean=ocean)  # the highest 35% are declared hills
+    ml = find_threshold_f(e, 0.10, ocean=ocean)  # the highest 10% are declared low mountains
+    mml = find_threshold_f(e, 0.06, ocean=ocean)  # the highest 10% are declared medium mountains
+    hml = find_threshold_f(e, 0.02, ocean=ocean)  # the highest 10% are declared high mountains
```
tcld: I added in values for the different mountain-levels. The percentages, however, are merely guessed so far. I could use some better suggestions on those ( @ftomassetti maybe? Also, @esampson seems to know things like these).

worldengine/draw.py
```diff
@@ -12,13 +11,6 @@
 ### For draw_satellite ###
 NOISE_RANGE = 15 # a random value between -NOISE_RANGE and NOISE_RANGE will be added to the rgb of each pixel
 
-# These are arbitrarily-chosen elevation cutoffs for 4 different height levels. 
-# Some color modifiers will be applied at each level
-HIGH_MOUNTAIN_ELEV = 215
-MOUNTAIN_ELEV      = 175
-HIGH_HILL_ELEV     = 160
-HILL_ELEV          = 145
-
```
tcld: These are the variables that are supposed to be replaced by this PR.

worldengine/draw.py
```diff
@@ -277,74 +275,42 @@ def get_biome_color_based_on_elevation(world, elev, x, y):
...
+        # Average the biome's color with the MOUNTAIN_COLOR to tint the terrain
+        biome_color = average_colors(biome_color, MOUNTAIN_COLOR)
+
+    ####### Case 3 - elevation is somewhat high ########
+    elif world.is_low_mountain((x, y)):
+        noise = add_colors(noise, HIGH_HILL_NOISE_MODIFIER)
+
+    ####### Case 4 - elevation is a little bit high ########
+    elif world.is_hill((x, y)):
+        noise = add_colors(noise, HILL_NOISE_MODIFIER)
```
tcld: All the static height-checks have been replaced by calls to the appropriate functions; output currently looks different than before but I suspect that is due to the parameters guessed in ```generation.py```.

Here the output of a test-world:
satellite_28070

worldengine/generation.py
tcld: As a note: Neither these values nor the corresponding functions in ```World.py``` have really been used before (maybe in one or two places). So this shouldn't cause terribly broken code and might offer some opportunities for future improvements to some of the generation-algorithms.

tests/draw_test.py
```diff
@@ -102,70 +85,77 @@ def test_elevation_color(self):
...
 
     def test_draw_scatter_plot(self):
         w = World.open_protobuf("%s/seed_28070.world" % self.tests_data_dir)
-        target = PixelCollector(16, 16)
-        draw_scatter_plot(w, 16, target)
+        target = PNGWriter.rgba_from_dimensions(512, 512, filename=None)
+        draw_scatter_plot(w, 512, target)
+        self._assert_img_equal("scatter_28070", target)
+
+    def test_draw_satellite(self):
```
ftomassetti: new tests: gooooood!

worldengine/draw.py
```diff
@@ -277,74 +275,42 @@ def get_biome_color_based_on_elevation(world, elev, x, y):
...
-
     return this_tile_color
+
+
 # ----------------------
 # Draw on generic target
 # ----------------------
 
-
-class ImagePixelSetter(object):
```
ftomassetti: Why we do not need this at all anymore? Do we still have functions to draw which works on an image and on the screen?

worldengine/draw.py
```diff
@@ -552,7 +517,7 @@ def draw_precipitation(world, target, black_and_white=False):
         low = world.precipitation['data'].min()
         high = world.precipitation['data'].max()
         floor = 0
-        ceiling = 255
+        ceiling = 255  # could be changed into 16 Bit grayscale easily
```
ftomassetti: do we want to use a constant then? Something like "max value"?

worldengine/drawing_functions.py
```diff
@@ -40,11 +40,11 @@ def draw_rivers_on_image(world, target, factor=1):
 
     for y in range(world.height):
         for x in range(world.width):
-            if world.is_land((x, y)) and (world.river_map[x, y] > 0.0):
+            if world.is_land((x, y)) and (world.river_map[y, x] > 0.0):
```
ftomassetti: I think river_map and lake_map are inverted w.r.t. to the other arrays

worldengine/generation.py
```diff
@@ -124,12 +124,17 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
     """
     e = world.elevation['data']
     ocean = fill_ocean(e, ocean_level)
-    hl = find_threshold_f(e, 0.10)
-    ml = find_threshold_f(e, 0.03)
+    pl = find_threshold_f(e, 0.70, ocean=ocean)  # the highest 70% of land are declared plains
+    hl = find_threshold_f(e, 0.35, ocean=ocean)  # the highest 35% are declared hills
+    ml = find_threshold_f(e, 0.10, ocean=ocean)  # the highest 10% are declared low mountains
+    mml = find_threshold_f(e, 0.06, ocean=ocean)  # the highest 10% are declared medium mountains
```
ftomassetti: we should change the percentage in the comment (or remove the percentage from all the comments)

worldengine/image_io.py
```diff
@@ -0,0 +1,227 @@
+"""
+This file is supposed to be the wrapper around image-processing modules like
```
ftomassetti: It seems a good idea to me. We should also start to structure modules

worldengine/world.py
```diff
@@ -417,66 +422,55 @@ def tiles_around_many(self, pos_list, radius=1, predicate=None):
...
-        else:
-            mi = 1
-        mountain_level = self.elevation['thresholds'][mi][1]
+        high_mountain_level = self.elevation['thresholds'][5][1]
         x, y = pos
-        return self.elevation['data'][y, x] > mountain_level + 4.0
+        return high_mountain_level <= self.elevation['data'][y, x]
 
-    def is_hill(self, pos):
+    def level_of_mountain(self, pos):
```
ftomassetti: this one perhaps needs a comment

worldengine/draw.py
tcld: This is due to PR #163 . The functions for drawing haven't really been changed; the image-handler is now in ```image_io.py``` - a reader and a writer.

worldengine/draw.py
tcld: The used image-writer has a function to retrieve the maximum color-depth (```get_max_colors()```, I believe); but since this function could be called by UI-functions they would have to provide a function like that, too. (Although it'll always return 255 in that case, I assume.) So I didn't touch the constant value for now.

worldengine/drawing_functions.py
tcld: Was there a reason for that? I changed it in PR #163, as part of unifying the coordinate-usages. It was not just because this inconsistency was bugging me, but when I replaced the image-writer the ancient-maps were a little bit messed up due to the inconsistent coordinates. Instead of working on a workaround I flipped the arrays. (Which was fairly easy, that's why all of this code uses [y, x]-notation. The other way would have been a lot more work.)

worldengine/generation.py
tcld: You are right. I will do that once the values are optimized a bit. I will have to touch those lines anyway.

worldengine/world.py
tcld: Any suggestion? "calculates height of the mount relative to mountain-base"?

tcld:
I rebased this and got rid of all conflicts. The data corresponding to this PR can be found [here](https://github.com/tcld/worldengine-data/commit/09c9150c68ed82580421ec36b268a403bdbc1865).

As a first draft, this is done. But it certainly needs more fine-tuning, otherwise the data would probably not look that different from before.

worldengine/World.proto
```diff
@@ -57,62 +57,65 @@ message World {
...
-    required double heightMapTh_sea      = 7;
-    required double heightMapTh_plain    = 8;
-    required double heightMapTh_hill     = 9;
+    required DoubleMatrix heightMapData       = 6;
+    required double heightMapTh_sea           = 7;
+    required double heightMapTh_plain         = 8;
+    required double heightMapTh_hill          = 9;
+    required double heightMapTh_low_mountain  = 10;
+    required double heightMapTh_med_mountain  = 11;
+    required double heightMapTh_high_mountain = 12;
```
tcld: Is it problematic to add the new variables in the "middle" instead of the end?

worldengine/generation.py
```diff
@@ -125,12 +125,17 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
     """
     e = world.elevation['data']
     ocean = fill_ocean(e, ocean_level)
-    hl = find_threshold_f(e, 0.10)  # the highest 10% of all (!) land are declared hills
-    ml = find_threshold_f(e, 0.03)  # the highest 3% are declared mountains
+    pl = find_threshold_f(e, 0.70, ocean=ocean)  # the highest x% of land are declared plains
+    hl = find_threshold_f(e, 0.35, ocean=ocean)  # the highest x% are declared hills
+    ml = find_threshold_f(e, 0.10, ocean=ocean)  # the highest x% are declared low mountains
+    mml = find_threshold_f(e, 0.06, ocean=ocean)  # the highest x% are declared medium mountains
+    hml = find_threshold_f(e, 0.02, ocean=ocean)  # the highest x% are declared high mountains
```
tcld: I could certainly use help determine proper values here.

worldengine/generation.py
```diff
@@ -105,12 +105,20 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
...
-    hl = find_threshold_f(e, 0.10)  # the highest 10% of all (!) land are declared hills
-    ml = find_threshold_f(e, 0.03)  # the highest 3% are declared mountains
+
+    # values chosen to emulate the exponential behavior seen here:
+    # http://www.ngdc.noaa.gov/mgg/global/etopo1_surface_histogram.html
+    pl = find_threshold_f(e, 0.95, ocean=ocean)  # the highest x% of land are declared plains; save some for beaches (?)
+    hl = find_threshold_f(e, 0.50, ocean=ocean)  # the highest x% are declared hills
+    ml = find_threshold_f(e, 0.25, ocean=ocean)  # the highest x% are declared low mountains
+    mml = find_threshold_f(e, 0.12, ocean=ocean)  # the highest x% are declared medium mountains
+    hml = find_threshold_f(e, 0.06, ocean=ocean)  # the highest x% are declared high mountains
```
tcld: I picked these values after inspecting the results of #203 and [Earths hypsographic curve](http://www.ngdc.noaa.gov/mgg/global/etopo1_surface_histogram.html). The real world seems to follow something like 2^x according to that graph.

worldengine/simulations/erosion.py
```diff
@@ -152,9 +152,7 @@ def river_sources(world, water_flow, water_path):
                 while not neighbour_seed_found:
 
                     # have we found a seed?
-                    if world.is_mountain((cx, cy)) and \
-                            water_flow[cy, cx] >= RIVER_TH:
-
+                    if world.is_low_mountain((cx, cy), atleast=True) and water_flow[cy, cx] >= RIVER_TH:
```
tcld: I decided that rivers shouldn't start on hills anymore. Higher mountains would be an option as well.

worldengine/simulations/temperature.py
```diff
@@ -89,13 +92,8 @@ def _calculate(world, seed, elevation, mountain_level):
...
-                    if elevation[y, x] > (mountain_level + 29):
-                        altitude_factor = 0.033
-                    else:
-                        altitude_factor = 1.00 - (
-                            float(elevation[y, x] - mountain_level) / 30)
-                    t *= altitude_factor
-                temp[y, x] = t
+
+                # vary temperature based on height
+                temp[y, x] = t * numpy.interp(e[y, x], [mountain_level, highest_mountain_level], [1.00, min_altitude_factor])
```
tcld: Does basically the same as the old code, but there are not arbitrary constants used; instead only heights are taken into account. `min_altitude_factor` could later be made global.

worldengine/world.py
```diff
@@ -416,66 +421,67 @@ def tiles_around_many(self, pos_list, radius=1, predicate=None):
...
-            mi = 1
-        mountain_level = self.elevation['thresholds'][mi][1]
+        mountain_level = self.elevation['thresholds'][2][1]  # hills are mountains, too
         x, y = pos
-        return self.elevation['data'][y][x] > mountain_level
+        return mountain_level <= self.elevation['data'][y, x]
 
-    def is_low_mountain(self, pos):
-        if not self.is_mountain(pos):
+    def is_hill(self, pos, atleast=False):
```
tcld: I chose to add an `atleast` parameter instead of adding another four functions that basically contain copied code.

tcld:
I updated this one to make it use values that are close to Earth's. I am now satisfied with the results. :)

ftomassetti:
Restarting tests

worldengine/World.proto
ftomassetti: well, that breaks compatibility but in this case I would say it is intended

worldengine/generation.py
```diff
@@ -105,12 +105,20 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
     """
     e = world.elevation['data']
     ocean = fill_ocean(e, ocean_level)
-    hl = find_threshold_f(e, 0.10)  # the highest 10% of all (!) land are declared hills
-    ml = find_threshold_f(e, 0.03)  # the highest 3% are declared mountains
+
+    # values chosen to emulate the exponential behavior seen here:
+    # http://www.ngdc.noaa.gov/mgg/global/etopo1_surface_histogram.html
+    pl = find_threshold_f(e, 0.95, ocean=ocean)  # the highest x% of land are declared plains; save some for beaches (?)
+    hl = find_threshold_f(e, 0.50, ocean=ocean)  # the highest x% are declared hills
```
ftomassetti: 50% isn't a lot? It is based on real data?

tcld: 50% of only the land. "Hills" to me sounds like those small bumps in the world - 50% sounded reasonable. What would you suggest?
I basically tried to emulate this:
hypsographic curve Earth
And that seemed like 2^x to me.

tcld:
Also: There is data that comes with this, https://github.com/Mindwerks/worldengine-data/pull/17.
Most of the output slightly changes due to the change to the mountain-levels.

worldengine/generation.py
tcld: Here a plot of how the levels look:
plot

ftomassetti: I think we could use the definition of hill as 300m of elevation. I am not sure if it is w.r.t. to the surrounding terrain or to the level of the sea. I am looking for more precise data. to understand better

tcld: Looking at the two images above, the scale of the platec-world to Earth seems to be roughly 1000:1. So hills right now are at roughly 1000m - the value would be lower if the first part of the graph wasn't as steep as it is.
You said 50% seemed like a lot. Moving hills *down* will increase that number even further - to maybe 75%.

The graphs can probably be used to find reasonable values. Although the definitions of hills, mountains etc. are kind of arbitrary at the moment.

ftomassetti: Hill used to represent the top 10% of all the world, if land is 1/3 of the world it means they used to ~33% of land, right? So that mean we are getting 50% more hills than before. Mountains should roughly stay the same instead (if I am not wrong)

tcld: Actually my former comment in the code was off: *Plains* used to be the top 10% of the world, hills only the top 3%. The current value pushes that to 15%. So if we wanted less, we should pick a base other than 2...so more like 10^x? That sounds very drastic to me.

Don't forget, though, that hills were the highest portion of the world before the new levels were added. From that point of view it might be perfectly fine to have 50% hills and maybe equate the new medium mountains with the former hills.

tcld: The first two plots in https://github.com/Mindwerks/worldengine/pull/203 show the difference of the values pretty nicely (old levels first, then new levels). The change is not as drastic as it might seem.

tcld: I just checked something: Look at [the first image](https://github.com/Mindwerks/worldengine-data/pull/17/files) in swipe-mode.
Doesn't really tell anything about the quality of the chosen values, only that I should check if the ancient maps need a little tweak to make use of the new levels.

The values themselves seem pretty good to me; the old ones seem much more sketchy, with hills being the substitute for a proper mountain-level. Once we agree on good values, I will check the three places that actually make use of them (river-generation, temperature-generation and ancient map-generation). I think there is some detail to be gained in those. :)

worldengine/generation.py
```diff
@@ -105,12 +105,20 @@ def initialize_ocean_and_thresholds(world, ocean_level=1.0):
...
-    hl = find_threshold_f(e, 0.10)  # the highest 10% of all (!) land are declared hills
-    ml = find_threshold_f(e, 0.03)  # the highest 3% are declared mountains
+
+    # values chosen to emulate the exponential behavior seen here:
+    # http://www.ngdc.noaa.gov/mgg/global/etopo1_surface_histogram.html
+    pl = find_threshold_f(e, 0.80, ocean=ocean)  # the highest x% of land are declared plains; save some for beaches (?)
+    hl = find_threshold_f(e, 0.40, ocean=ocean)  # the highest x% are declared hills
+    ml = find_threshold_f(e, 0.20, ocean=ocean)  # the highest x% are declared low mountains
+    mml = find_threshold_f(e, 0.10, ocean=ocean)  # the highest x% are declared medium mountains
+    hml = find_threshold_f(e, 0.05, ocean=ocean)  # the highest x% are declared high mountains
```
tcld: I changed the values a bit, created new plots and want to collect all the important points in a single post:
The former "hills"-level was assigned to the highest 3% of the world, the following 7% (for a total of 10%) was declared plains.

The new values declare the following values, in relation to *total* land (in the code the values are relative to non-submerged land):
plains ~28%, hills 14%, low mountains 7%, medium mountains 3.5% and high mountains 1.8%.

So the former "hills" roughly correspond to "medium mountains" now. Thus the changes to river-, temperature- and ancient map-generation. The choices in there made sense, but the hill-level was very high.

The following two plots show the old and new height-levels:
seed_13051_hypsographic
seed_13051_hypsographic

The plains are set to 80% since due to possible depressions in the land 100% could end up below sea-level. 80% leaves a little wiggle-room.
Everything else is chosen to follow a rough 2^x shape, as I estimated it from here (supposedly real data): hypsographic curve Earth

If anybody has more accurate data, please point me to it (other criticism is welcome, too). For now the values seem reasonable and [the results of world-generation](https://github.com/Mindwerks/worldengine-data/pull/17/files) look good to me.

worldengine/simulations/temperature.py
```diff
@@ -27,13 +25,18 @@ def execute(self, world, seed):
...
         height = world.height
+        e = world.elevation['data']
+        mountain_level = world.start_mountain_th()  # returns minimum elevation value for mountains (i.e. hills)
+        highest_mountain_level = e.max() # world.level_of_mountain(numpy.unravel_index(e.argmax(), e.shape))  # height of the highest point relative to mountain_level
 
         rng = numpy.random.RandomState(seed)  # create our own random generator
         base = rng.randint(0, 4096)
         temp = numpy.zeros((height, width), dtype=float)
+        
+        min_altitude_factor = 0.3  # used for the top of the highest mountain; must be in [0, 1]
```
tcld: This factor makes the highest mountain tops pretty cold, the effects can be seen in [the temperature-map](https://github.com/Mindwerks/worldengine-data/pull/17/files). It is mostly a guess, consolidated by Googling the average temperature in Antarctica and on the Mount Everest.

PR #189, opened 2015-11-06
Frozen oceans on satellite map

tcld:
Here some examples. Based on PR #185, only the last commit is currently relevant.
@pangal Thanks for your awesome work! If you have an idea or just want to do this all yourself, just tell me. :)

Discussion regarding parameters or improved techniques welcome. Here some example images (bigger ones will follow):
seed_10138_satellite

Looking at the picture I think I should bring ```diff_freeze_limit``` and ```diff_freeze_chance``` closer together. Maybe it should even depend on the size of the map.


ftomassetti:
wow! So much cool stuff is going on in this project!

pangal:
This looks awesome! Thanks for the additions @tcld !

tcld:
Happy to hear you like it. :)

I am not at all satisfied yet, though. The noise is too much, maybe there should be a modification of the chance-to-freeze depending on how many of the surrounding tiles are frozen/solid.

tcld:
After a change in the code I found this one:
seed_25095_satellite
It is obviously not orbiting its star at a 90-degree angle. =)

seed_25862_satellite
These two look better, I think. I'll see how the bigger maps turn out.

tcld:
And after some "finishing" touches (no worries, these are the last images :P ):
seed_3543_satellite

Heavily tilted against the plain of rotation.
Kind of a cold planet, maybe far away from the sun. (Could be other effects, too.)

Surrounding tiles are now taken into account and some of the parameters aren't fixed but instead calculated now.
The style of the code is kind of bad, I think. But for a first draft it's good enough. :)

psi29a:
Is the frozen ocean wrapping around to the other side (along the x-axis)?

If not, can we make it so?  So that if we wrap around on a cylinder it looks seamless.

tcld:
No, it isn't wrapping around. How would I do that properly? (I think I saw something like that in ```temperature.py```, right?)

psi29a:
You can 'cheat' by doing a final pass that blends/blurs the frozen-ocean pixels together.

tcld:
Wait a moment...if ```temperature.py``` actually wraps around (and [it does](https://github.com/Mindwerks/worldengine/blob/master/worldengine/simulations/temperature.py#L84) ) then the ice should wrap around automatically since the water freezes wherever the ocean is cold enough.

tcld:
I checked, it does *not* nicely wrap. But it's close.

psi29a:
I suspect you are right, fix that and we fix biomes too. ;)

pangal:
Seems to me like maybe we should have an "icecap" or "ice" biome, which is set at some point earlier in world generation? Conceptually, the `draw_satellite` function should probably be drawing things that exist in the world's data structure, not necessarily adding new things. 

Additionally, certain projects using worldengine may wish to use those tiles as traversable "land" tiles. If the icecaps are set earlier can show up on some of the other map outputs, and it can be saved into the world structure too. That way, in the satellite view map we can just handle the coloring.

Either way, I think the new icecaps look great!

tcld:
The ice-code could be put into it's own function (somewhere in the simulations-folder) and create a boolean map that is set to True wherever there is frozen ocean. ```draw_satellite()``` would then just need to add some color according to the ice-map.

That's probably the better solution since the icy areas shown above are indeed nothing but visual effects.

esampson:
@psi29a Biomes should already be wrapping properly. The temperature and humidity maps both wrap and they're what determines the biomes. There's no additional noise or random pattern on top of that.

Strangely there were a couple of temperature maps included in this conversation earlier and the first one did not seem to be wrapping around, but I just checked with a temperature map on my system. Were those images cropped? They seem to be at an odd resolution and the first one seems to have the temperature bands shifted. The polar band on the bottom seems very small.

tcld:
Those temperature-maps were an accident, I dragged in the wrong images.

Maybe I didn't properly handle the orbital-parameter implementation and some of the temperature-maps aren't nicely wrapping? (The orbital parameters are by the way what makes some polar bands really small/large.)

tcld:
Here are the temperature-maps for the last three worlds (and the generation-parameters...which will probably be useless due to the changes in this PR).
seed_3543_temperature
```
Seed           : 3543
Dimensions     : 960x600
Plates         : 8
Ocean          : 65.00%
```

seed_15622_temperature
```
Seed           : 15622
Dimensions     : 480x360
Plates         : 13
Ocean          : 65.00%
```

seed_22638_temperature
```
Seed           : 22638
Dimensions     : 480x360
Plates         : 4
Ocean          : 65.00%
```

esampson:
Ah. Ok. You've got code in there that's not in the master. My guess is that that explains why you aren't wrapping properly.

How would orbital parameters shift the equatorial zones? I can see orbital parameters affecting how hot/cold things are and temperature swings throughout the year but unless a planet is tidally locked it seems like each pole spends about half the year pointed to/away from the star (and a tidally locked planet is going to be completely different from what we are seeing).

tcld:
The zones would slowly shift for a tilted planet; since we aren't simulating time I decided to look at it as if that shift wouldn't be happening. So the "heat-band" moves north and south a little bit, depending on the seed.

The code in here doesn't affect the temperature-map itself, it is mostly about the image-output.

esampson:
Ah! Yes, they would shift back and forth over the year, but I'm pretty sure our maps are suppose to represent year averages, not current temperatures. Biomes simply don't shift that fast.

psi29a:
I agree with @esampson

esampson:
seed_15622_temperature

Interesting. This doesn't seem to be wrapping quite properly, either. I wonder if it is because of the 4:3 ratio.

tcld:
The temperature-maps strangely show very little structure; I didn't notice until @psi29a pointed it out somewhere (much older temperature-maps looked more like the precipitation-maps now). When I first looked at several dozens of WorldEngine-worlds, I only noticed that almost all temperature-maps looked the same - and I wanted to see some variety. The orbital parameters might not be 100% properly simulated but there are some corners to be cut with every part of the simulation. (I much prefer the slightly shifting band.)

@esampson The aspect ratio is an interesting idea.

tcld:
Here an old temperature-map:
https://github.com/tcld/worldengine-data/commit/f45fe81e107fa3afef8bd8436b78311795653593

esampson:
Lack of variety is probably because it is tuned to produce a fairly 'real' result. If you look at the temperature.py routine you'll see that there's a variable that weighs between totally random noise and pure latitude/altitude based temperature.

I had forgotten about it and the fact that I wanted to make that a command line parameter as well.

esampson:
Hmm. Moving to a 2:1 ratio doesn't seem to be fixing it either. Maybe its because it isn't powers of 2? I'll have to look at it in a bit.

ftomassetti:
_Back in the old times temperature maps used to be different my dear..._

tcld:
@esampson: Did you ever find out why the temperature-maps don't always fully wrap?

Once the parent-PRs are polished, I would like to move the ice-cap generation into a new module inside the simulations-folder. That would make it necessary to add a new world-"variable" (a boolean map named maybe ```ice_map```) that can later be drawn by the satellite-method (and maybe made use of by other parts of the simulation). Would that be ok, @ftomassetti / @psi29a ? Or does the use of that kind of map seem too limited to store it with the data? (A change like that would also make it necessary to update the save-routines a bit, including #153 .)

ftomassetti:
I am in favor of adding the variables: we want WorldEngine to be used as a library and users could take advantage of information about the frozen ocean. Of course many could just ignore it but that is fine.

tcld:
Ok, I'll add the new variable then. Once #163 and then #185 are out of the way, I'll move the code around and finish this PR. If #153 were to be merged before this PR, I could probably add the new variable to the HDF5-saves myself.

tcld:
I finished pushing the code around. This is now technically done, I don't know enough about polar cap-formation in real life to make it any better. If anybody has any suggestions regarding the free parameters (or the whole algorithm), please notify me. I guessed them all by looking at the output and deciding whether I liked it or not.

Here the output of my latest test (the temperature band was quite far to the south so there is no ice there:
seed_9390_satellite
seed_9390_icecaps

I will add some comments to the code, there are a couple of places where things could be made prettier.

generate_protobuf_stubs.sh
```diff
@@ -1,2 +0,0 @@
-#!/bin/bash
-protoc -I=worldengine --python_out=worldengine/protobuf worldengine/World.proto
```
tcld: Removed according to https://github.com/Mindwerks/worldengine/issues/190.

worldengine/cli/main.py
```diff
@@ -543,7 +550,10 @@ def main():
             '%s/%s_scatter.png' % (args.output_dir, world_name))    
         if args.satelite_map:
             draw_satellite_map(world,
-            '%s/%s_satellite.png' % (args.output_dir, world_name))    
+            '%s/%s_satellite.png' % (args.output_dir, world_name))
+        if args.icecaps_map:
+            draw_icecaps_map(world,
+            '%s/%s_icecaps.png' % (args.output_dir, world_name))
```
tcld: I added an option to look at the height of the ice. Not too useful currently, but maybe it will be in the future.

worldengine/simulations/icecap.py
```diff
@@ -0,0 +1,79 @@
...
+        # local variables
+        icecap = numpy.zeros((world.height, world.width), dtype=float)
+        rng = numpy.random.RandomState(seed)  # create our own random generator
+
+        # map that is True whenever there is land or (certain) ice around
+        solid_map = numpy.logical_or(temperature <= freeze_chance_threshold + temp_min, numpy.logical_not(ocean))
+
+        for y in range(world.height):
+            for x in range(world.width):
+                if ocean[y, x]:  # or world.river_map[y, x] > 0 or world.lake_map[y, x] > 0 or world.watermap['data'][y, x] > 0:
```
tcld: There could be more checks, the code itself is not dependent on there being ocean. But I wouldn't know how to properly handle other freezable places.

worldengine/simulations/icecap.py
```diff
@@ -0,0 +1,79 @@
...
+        freeze_chance_window = 0.20  # the warmest x% of freezable area won't completely freeze (RNG decides)
+        surrounding_tile_influence = 0.5  # chance-modifier to freeze a slightly warm tile when neighbors are frozen
+
+        # secondary constants
+        temp_min = temperature.min()  # coldest spot in the world
+        freeze_threshold = world.temperature['thresholds'][0][1]  # upper temperature-limit for freezing effects
+        # Cold biomes: TODO: find and pick most appropriate threshold
+        #    polar: self.temperature['thresholds'][0][1]
+        #   alpine: self.temperature['thresholds'][1][1]
+        #   boreal: self.temperature['thresholds'][2][1]
```
tcld: I wasn't sure which temperature-boundary to pick. They are all cold enough to make water permanently freeze, I believe. And even "boreal" didn't create insane amounts of ice (due to ```max_freeze_percentage```).

tcld:
Updated for HDF5.
Data for this PR can be seen here: https://github.com/tcld/worldengine-data/commit/14b9c6937644e419d98a4f78e825a17fba5515d0
I will create a PR for the data once #185 is merged so the order isn't mixed up (if that's even possible).

tcld:
pyflakes is complaining about a line of code in `World_pb2.py`, which is generated by Protobuf. Any idea what to do about that aside from manually removing the line?

Also, even though the problems would vanish with new data, the Travis-tests are mostly failing due to a variable having an unknown value. Does that mean that, whenever a variable is added to `World`, saves will stop to be compatible at all?

worldengine/World.proto
```diff
@@ -1,3 +1,4 @@
+syntax = "proto2";
```
tcld: Would it be advisable to switch to "proto3"?

psi29a: Yes, I've always commented/deleted that one line. It is automatically produced, but never consumed and pyflakes chokes on it.

psi29a: Here is the API/docs for proto2 and proto3.  Proto3 just extends proto2 and should be backwards compatible.

tcld: `World.proto` has to be changed a bit to make it work for "proto3", I tried. I also found the documentation. ;)
So should I rather use "proto3" or is that something to be left for the future? I assume that the save-files won't be compatible between proto2 and 3 (the result is different according to git).

tcld:
@psi29a Could you restart the tests for this, please?

tcld:
I generated `World_pb2.py` with Protobuf 3.0, the failed test indicates version 2.6.1. I guess that's a problem, there most likely was no need to use 'syntax="proto2"' in Protobuf 2.

tcld:
I changed the Protobuf-version for the tests, now everything is fine. :)

psi29a:
awesome...

PR #192, opened 2015-11-13
Removed (almost) every last trace of support for pickle

tcld:
Since I wanted to [add a new variable to the World](https://github.com/Mindwerks/worldengine/pull/189) I noticed that I couldn't regenerate the pickle test-worlds after https://github.com/Mindwerks/worldengine/pull/187 had been merged (I did it via a bash-script before, but WorldEngine didn't take a parameter to output a pickle-world anymore).
After a bit of thinking I noticed that "load-only support" for pickle is worthless once new variables are added to the world since compatibility will most likely be lost anyway; @psi29a [seemed to agree](https://github.com/Mindwerks/worldengine/pull/187) that pickle-support could probably be purged completely.

So what this does might seem a lot:
* completely remove pickle-support
* update ```data_generator.py``` so it actually outputs all necessary data (should have done that earlier)
* also updated the worlds data to use a single world-file with seed 28070 instead of four different files (I didn't see the purpose; comments on that are welcome)
* removed two tests that seemed useless now

**EDIT:** [Data for this PR](https://github.com/Mindwerks/worldengine-data/pull/13)

### DIFF
--- a/tests/data/data_generator.py
+++ b/tests/data/data_generator.py
@@ -8,13 +8,18 @@ because the plate simulation steps do not provide the same results on all the pl
 """
 
 import os
-import platform
-from worldengine.plates import _plates_simulation
+from worldengine.plates import world_gen, _plates_simulation
 
 
 def main(tests_data_dir):
-    w = _plates_simulation("Foo", 300, 200, 279)
-    w.to_pickle_file("%s/py%s_plates_279.world" % (tests_data_dir, platform.python_version_tuple()[0]))
+    w = world_gen("seed_28070", 300, 200, 28070)
+    w.protobuf_to_file("%s/seed_28070.world" % tests_data_dir)
+
+    # w = world_gen("seed_48956", 450, 300, 48956)
+    # w.protobuf_to_file("%s/seed_48956.world" % tests_data_dir)
+
+    # w = world_gen("biome_test", 128, 128, 6908)
+    # w.protobuf_to_file("%s/biome_test.world" % tests_data_dir)
 
 
 if __name__ == '__main__':
--- a/tests/data/data_generator.py
+++ b/tests/data/data_generator.py
@@ -15,12 +15,6 @@ def main(tests_data_dir):
     w = world_gen("seed_28070", 300, 200, 28070)
     w.protobuf_to_file("%s/seed_28070.world" % tests_data_dir)
 
-    # w = world_gen("seed_48956", 450, 300, 48956)
-    # w.protobuf_to_file("%s/seed_48956.world" % tests_data_dir)
-
-    # w = world_gen("biome_test", 128, 128, 6908)
-    # w.protobuf_to_file("%s/biome_test.world" % tests_data_dir)
-
 
 if __name__ == '__main__':
     blessed_images_dir = os.path.dirname(os.path.realpath(__file__))
--- a/tests/drawing_functions_test.py
+++ b/tests/drawing_functions_test.py
@@ -1,5 +1,4 @@
 import unittest
-import platform
 
 from worldengine.drawing_functions import draw_ancientmap, gradient, draw_rivers_on_image
 from worldengine.world import World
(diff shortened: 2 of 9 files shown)


tcld:
I would prefer if https://github.com/Mindwerks/worldengine/pull/185 could be merged before this one, so I don't have to regenerated the data for that PR.
My other PRs can't be merged before this one since I wouldn't be able to generate the data needed for them. Pickle has to go first (or somehow fixed so that test-worlds can be generated again).

tcld:
The tests are passed, of course. But the data should still be updated to at least remove the now unneeded binary files.

tests/data/data_generator.py
```diff
@@ -8,13 +8,18 @@
...
-    w = _plates_simulation("Foo", 300, 200, 279)
-    w.to_pickle_file("%s/py%s_plates_279.world" % (tests_data_dir, platform.python_version_tuple()[0]))
+    w = world_gen("seed_28070", 300, 200, 28070)
+    w.protobuf_to_file("%s/seed_28070.world" % tests_data_dir)
+
+    # w = world_gen("seed_48956", 450, 300, 48956)
+    # w.protobuf_to_file("%s/seed_48956.world" % tests_data_dir)
+
+    # w = world_gen("biome_test", 128, 128, 6908)
+    # w.protobuf_to_file("%s/biome_test.world" % tests_data_dir)
```
psi29a: why comments and not just removed?

tcld: I started adding that code so `data_generator.py` would be able to generate all data at once and I wouldn't have to use an external script. Then I starting removing references to all but one world and I wasn't sure if these specific worlds would maybe be needed again in case the switch to a single world was a bad idea. If everything else is fine, I guess I should remove the commented code.

worldengine/cli/main.py
```diff
@@ -186,32 +183,18 @@ def __seems_protobuf_worldfile__(world_filename):
...
     pb = __seems_protobuf_worldfile__(world_filename)
-    pi = __seems_pickle_file__(world_filename)
-    if pb and pi:
-        print("we cannot distinguish if the file is a pickle or a protobuf "
-              "world file. Trying to load first as protobuf then as pickle "
-              "file")
+    if pb:
+        # print("we cannot distinguish if the file is a pickle or a protobuf "
+        #       "world file. Trying to load first as protobuf then as pickle "
+        #       "file")
```
tcld: Another part of code where I wasn't sure whether I should permanently remove it. Maybe other formats will have to be checked for in the future within this function and it might be a useful little reminder? Maybe it's completely useless, though.

tcld:
Comments removed. (Since the data has already been merged, this should probably be merged soon, too.)

tests/drawing_functions_test.py
```diff
@@ -13,13 +12,7 @@ def setUp(self):
...
         self.w = World.open_protobuf("%s/seed_28070.world" % self.tests_data_dir)
 
-    def test_draw_ancient_map_factor1(self):
-        w_large = World.from_pickle_file("%s/py%s_seed_48956.world" % (self.tests_data_dir, platform.python_version_tuple()[0]))
-        target = PNGWriter.rgba_from_dimensions(w_large.width, w_large.height)
-        draw_ancientmap(w_large, target, resize_factor=1)
-        self._assert_img_equal("ancientmap_48956", target)
-
-    def test_draw_ancient_map_factor3(self):
+    def test_draw_ancient_map(self):
```
ftomassetti: We could possibly want to have the two tests in the future but that is really not urgent

ftomassetti:
LGTM, I will award it will the "ship-it-squirrel" :shipit:

ftomassetti:
Let's wait for tests to pass and merge this

tests/drawing_functions_test.py
tcld: One was testing a pickle-file, one a protobuf-file. They did pretty much the same anyway. Should another format need a test like this, maybe it should get it's own `drawing_functions_test.py`?

tcld:
What's a "shipt-it-squirrel"? Maybe I should have been more clear about the order in which things should be merged?

Tests will pass within the next two or three minutes, I guess. Already almost done.

ftomassetti:
It is something that is used at GitHub to show support for a PR (https://www.quora.com/GitHub/What-is-the-significance-of-the-Ship-It-squirrel).

Do no worry about that! Having the separate repositories for code and data these things happen and it will be solved soon.

tcld:
Ah, ok. (I think the tests just restarted. I was looking at the log of the almost finished tests, now it is back to square one. So...another at least ten minutes, I guess.)

ftomassetti:
Ok, tests are passing. Let's merge this one!

PR #197, opened 2015-11-16
Satellite-view as standard output; a bit of clean-up

tcld:
I wanted to make the satellite-view a standard output; while working on it I cleaned up very few lines of code.

I also noticed that generation including the new outputs (satellite and scatter) now takes roughly 50% longer than it took exactly four weeks ago. The satellite-view is probably slow, but I assume that something else changed.
The speed is still not bad, we are now at twice the speed of 0.18. But maybe somebody has an idea of what to improve.

### DIFF
--- a/worldengine/cli/main.py
+++ b/worldengine/cli/main.py
@@ -59,6 +59,7 @@ def generate_world(world_name, width, height, seed, num_plates, output_dir,
         filename = '%s/%s_precipitation.png' % (output_dir, world_name)
         draw_precipitation_on_file(w, filename, black_and_white)
         print("* precipitation image generated in '%s'" % filename)
+
         filename = '%s/%s_temperature.png' % (output_dir, world_name)
         draw_temperature_levels_on_file(w, filename, black_and_white)
         print("* temperature image generated in '%s'" % filename)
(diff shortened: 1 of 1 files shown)


psi29a:
What do you mean that we are double the speed of 0.18? Is that faster or slower, the direction is clear in your post. ;)

Do we really want to output everything by default? ;)

tcld:
I only made the satellite-view a default, and I don't think that's debatable.^^

"Double the speed" == "faster" ;P
Generation for me now takes about half as much time as in 0.18, including several of the optional outputs. But four weeks ago, I only needed 66% of the time I need now. So things got a little slower again.

tcld:
Tests here should be fine after #189 are merged.

psi29a:
This branch has conflicts that must be resolved

tcld:
Conflicts gone.

ftomassetti:
I think we should consider what images to generate automatically and which not. It makes probably more sense to generate the satellite view by default and disable temperature, precipitation and ocean images.
We could then add operations to generate specific images from the world file.
I would suggest we retarget this to the next release.

tcld:
Most of the images take about a second to be written, I am not sure if we wouldn't just be overcomplicating things by adding in more options etc.

psi29a:
We are changing the default behaviour, I think this PR can wait until after 0.19 release. :)

This is a library that also happens to have a CLI, as a CLI was should be able to show off as many features of the library as possible.

The GUI on the other hand can be stream-lined... however we need someone who has experience in UX to help design it. Or we just imitate/emulate others. ;)

tcld:
What do you mean by 'default behaviour' here?
How about renaming `--sat` to `--no-sat` instead (and setting the corresponding default value to `True`)? Would that be acceptable?

If you are talking about worldengine-gui, did anybody try to use it lately? I haven't, but it would probably be a good idea to see if it agrees with all the recent changes to worldengine.

psi29a:
I was just giving an example about the GUI. It should still work as-is, since it was bare-bones to begin with.

tcld:
I will close this for now. The cleanup-part has been snuck into one of my other PRs, and making the satellite a standard output could be easily done any time while also being slightly debatable due to the time it takes to be generated.

Issue #193, opened 2015-11-15
Ensure all features are documented in the manual and in the README


ftomassetti:

ftomassetti:
In the README we state "Libjpeg, libtiff and zlib are required by Pillow"
Should we remove this?

psi29a:
I think these are no longer necessary.  However the number of libs required by GDAL is immense... we'll have to deal with that later. ;)

PR #198, opened 2015-11-16
Preparing documentation for release v0.19.0

ftomassetti:
This is to address #193


README.md
```diff
@@ -324,10 +285,10 @@ If you are using WorldEngine please let us know!
 Contributors
 ============
 
-The merged project is maintained by [Bret Curtis](https://github.com/psi29a) and [Federico Tomassetti](https://github.com/ftomassetti).
+The this project is maintained by [Bret Curtis](https://github.com/psi29a) and [Federico Tomassetti](https://github.com/ftomassetti).
```
tcld: There is a superfluous word here (the/this).

manual/cli.rst
```diff
@@ -1,4 +1,100 @@
...
+|            | --recursion-limit=N  | you need that just if you encounter an error while generating very large maps                                                 |
++------------+----------------------+-------------------------------------------------------------------------------------------------------------------------------+
+| -v         | --verbose            | Enable verbose messages                                                                                                       |
++------------+----------------------+-------------------------------------------------------------------------------------------------------------------------------+
+|            | --version            | Display version information                                                                                                   |
++------------+----------------------+-------------------------------------------------------------------------------------------------------------------------------+
+| --bw       | --black-and-white    | Draw maps in black and white                                                                                                  |
```
tcld: I think this option only has an influence on precipitation- and temperature-map - so it is technically only useful for the "world" option.

manual/cli.rst
```diff
@@ -1,4 +1,100 @@
...
+
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+| Short     | Long                       | Description                                                                                          |
++===========+============================+======================================================================================================+
+| -r FILE   | --rivers=FILE              | produce a map of rivers, after the option it expects the name of the file where to generate the map  |
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+| --gs      | --grayscale-heightmap      | produce a grayscale heightmap                                                                        |
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+|           | --ocean_level=N            |  elevation cut off for sea level [default = 1.0]                                                     |
```
tcld: This option is still [not working](https://github.com/Mindwerks/worldengine/pull/134). Can that somehow be noted?

manual/cli.rst
```diff
@@ -1,4 +1,100 @@
...
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+|           | --temps #/#/#/#/#/#        |  Provide alternate ranges for temperatures. [default = .126/.235/.406/.561/.634/.876]                |
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+|           | --humidity #/#/#/#/#/#/#   | Provide alternate ranges for humidities. [default = .059/.222/.493/.764/.927/.986/.998]              |
```
tcld: Since this is the manual, these two could use some rudimentary explanation (must be in [0, 1]; names of the biomes in order of the given numbers and maybe even an explanation of the numbers themselves).
As is, this isn't comprehensible to me, and I should already know what these mean.

manual/cli.rst
```diff
@@ -1,4 +1,100 @@
...
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+| -gv N     | --gamma-value N            | N = Gamma value for temperature/precipitation gamma correction curve. [default = 1.25]               |
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+| -go N     | --gamma-offset N           | N = Adjustment value for temperature/precipitation gamma correction curve. [default = .2]            |
```
tcld: Can you maybe provide a link to be added here, @esampson? If I were to stumble over WorldEngine, I would want parameters to be understandable without using Google, at the very least.

manual/cli.rst
```diff
@@ -1,4 +1,100 @@
...
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+|           | ---not-fade-borders        | avoid fading borders                                                                                 |
```
tcld: Is this circumventing the smoothing of the border-areas before the ocean is "filled up"? If so, could it say something along those lines?

manual/cli.rst
```diff
@@ -1,4 +1,100 @@
...
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+|           | --scatter                  | generate scatter plot                                                                                |
```
tcld: An explanation or a link would be crucial in my eyes. It is *temperature vs. humidity*, I believe. So..."distribution of humidity in areas of the world plotted against their temperature" or something similar might be good.
And a link always helps, since certain areas of the plot should not be populated.

manual/cli.rst
```diff
@@ -1,4 +1,100 @@
...
++-----------+----------------------------+------------------------------------------------------------------------------------------------------+
+|           | --sat                      | generate satellite map                                                                               |
```
tcld: If there is a chance that https://github.com/Mindwerks/worldengine/pull/197 will be merged, this could be removed. The satellite-map is too pretty to be hidden behind a paramter.

manual/cli.rst
```diff
@@ -1,4 +1,100 @@
...
+
+Export options
+~~~~~~~~~~~~~~
+
++-----------+----------------------------+--------------------------------------------------------------------------------------------------------------------------------+
+| Short     | Long                       | Description                                                                                                                    |
++===========+============================+================================================================================================================================+
+|           | --export-format            | Export to a specific format such as BMP or PNG. See http://www.gdal.org/formats_list.html for possible formats [default = PNG] |
++-----------+----------------------------+--------------------------------------------------------------------------------------------------------------------------------+
+|           | --export-datatype          | Type of stored data, e.g. uint16, int32, float32 etc. [default = uint16]                                                       |
```
tcld: These two are my own mistakes: These parameters need to be like `--export-format='PNG'` etc. I obviously made a mistake, and now it looks as if the user wouldn't have to pass in anything.
I will try to update this - this part of the manual is probably not generated automatically, is it?

tcld:
I may have added a lot of comments. Some of the parameters really don't seem clear to me, although most of them certainly could be.

Is the explanation of parameters somehow auto-generated?

ftomassetti:
It is good to have many comments here. It is true that much stuff is rather obscure. I would try to make definitions more clear. In the long run it would be amazing to have examples on the effect of the single parameters.

For example showing the images generated with different parameters, different values, etc.

README.md
ftomassetti: thanks, solved

manual/cli.rst
ftomassetti: I added a note to specify that

manual/cli.rst
ftomassetti: well, here we document the intended behavior. We will fix it.... sooner or later :)

manual/cli.rst
ftomassetti: I agree this is not clear and I will revise it. However the space on the table is very limited. Ideally we should provide a longer explanation outside the table.

manual/cli.rst
tcld: I checked, and the parameter is indeed only used for temp- and precip-map. So I guess it would be appropriate to move it into the "Options valid only for generate"-group in `main.py` instead.

manual/cli.rst
ftomassetti: Ideally we want to show these things with images generated with the different options

manual/cli.rst
ftomassetti: Added "temperature vs. humidity". Also in this case I think that we should add examples and detailed explanation in the manual but outside the table

manual/cli.rst
ftomassetti: We could remove it in #197 :D

manual/cli.rst
ftomassetti: Added

manual/cli.rst
tcld: I guess it was supposed to include more plots than one, so the name was chosen to be very generic. But for now that description would be correct and helpful, I think.

manual/cli.rst
tcld: Depends on which is merged first. :P

ftomassetti: Good point, I should just click merge then :sunglasses: 
I am joking, we will see if #197 ends up in this release or in the next one and update the manual accordingly.

manual/index.rst
```diff
@@ -1,7 +1,7 @@
 WorldEngine - the best open-source world generator
```
tcld: How about "**an** open-source world generator"?^^

ftomassetti: Perhaps we can try to be bold :D

ftomassetti:
Mmmm why this one failed?
```

======================================================================
FAIL: test_draw_satellite (tests.draw_test.TestDraw)
----------------------------------------------------------------------
  File "/home/travis/build/Mindwerks/worldengine/tests/draw_test.py", line 161, in test_draw_satellite
    self._assert_img_equal("satellite_28070", target)
  File "/home/travis/build/Mindwerks/worldengine/tests/draw_test.py", line 56, in _assert_img_equal
    drawn_image.array[diff[0][0], diff[0][1]]))
AssertionError: Pixels at 0, 0 are different. Blessed [228 228 255 255], drawn [ 24  95 146 255]
----------------------------------------------------------------------
Ran 42 tests in 422.006s
```

tcld:
The problem only occurs for one of the Travis-thingies, looks like the code used for the test is not up-to-date - the blessed pixel shows ice (position and colour hint at it), the drawn pixel shows unfrozen ocean (so the icecap-PR is not yet in the code, it seems).

manual/index.rst
tcld: I like the change you made further down, going from "is intended to be the best open-source world generator" to "has the goal to be the best open-source world generator", it just sounds more humble. I would vote against boldness for the headline.^^

manual/cli.rst
esampson: Sorry, things came up that dragged me away for a bit.

I will try and put something together. Unfortunately the whole process is a little complicated. First I need to explain that the different biomes are chosen by the intersection of humidity and temperature values. Then I have to explain that both temperature and humidity and unitless values (because that's not mentioned anywhere). With those explained I can then explain how the numbers are used to map the unitless values into those ranges.

Unfortunately RST doesn't seem to support coloring the background of cells in any convenient way, which means the current biome chart I created will probably have to be handled as an image. Additionally I need to create an image of the Holdridge life zones that corresponds to our particular implementation.

ftomassetti:
Rebased and restarted the tests, let's see how it goes

manual/index.rst
ftomassetti: Well, the headline has to be shorter than "is intended to be the best open-source world generator".
I also searched many, many, many, many times for open-source world generator and I could not find any that is comparable to WorldEngine. If I can find one I would change the headline :D

tcld: How about this:
> WorldEngine - the best open-source world generator*
> ...
> *= Until proven otherwise. :P

ftomassetti: ah ah ah, it seems a disclaimer like you see in advertisement :)

PR #201, opened 2015-11-17
world.plates->numpy, some optimizations

tcld:
It seems I overlooked an array when making the other arrays numpy-arrays.
I stumbled over this after noticing that `center_world()`, for some reason, returned different results for Python 2 and 3 when I generated the following test-world. That is fixed with this, too.
```
width=1024
height=768
plates=3
ocean=0.71
seed=13051
```

Short summary of the changes in here:
* Overlooked world.plates when switching to numpy.
* Added world.plates to the serialization-tests (It had been stored before, but never was tested, I think.)
* Reworked center_land() to make use of numpy.
* Removed some superfluous functions/tests.

PS: Sorry for still doing stuff that is not related to an immediate release!


psi29a:
This actually looks like it should be included, because it falls under the mandate of 'use numpy, everywhere!'.

@ftomassetti wouldn't you agree?

tests/common_test.py
```diff
@@ -16,19 +16,6 @@ def test_get_and_set_verbose(self):
         set_verbose(False)
         self.assertEqual(False, get_verbose())
 
-    def test_matrix_min_and_max(self):
```
ftomassetti: Why we cancel the test but not the tested function?

tcld: I may have forgotten, will check that later.

tcld: Code removed.

tests/common_test.py
```diff
@@ -40,25 +26,6 @@ def test_counter(self):
         c.count("a")
         self.assertEqual("a : 1\nb : 3\n", c.to_str())
 
-    def test_rescale_value(self):
```
tcld: We can use `numpy.interp()` if something like `rescale_value()` is needed. As with many of the functions in `common.py`, it wasn't used anywhere anyway.

tcld:
Comments taken care of. I also removed another test + function since it wasn't used anywhere (and, like `matrix_min_and_max()`, had a numpy counterpart).

ftomassetti:
Let's merge #198 and then this one before changing release number and tagging the new release

PR #203, opened 2015-11-18
Hypsographic Curve

tcld:
While thinking about the values to choose for [PR 188](https://github.com/Mindwerks/worldengine/pull/188/files#diff-80eed145c1070172ffba1f642a13b731R108), I thought I might want to take a look at the elevation distribution of a world. So I added some code to generate a [hypsographic curve](http://www.ngdc.noaa.gov/mgg/global/etopo1_surface_histogram.html) of the world.

Data corresponding to this PR can be found [here](https://github.com/Mindwerks/worldengine-data/pull/16).

Example for the test-world (28070):
satellite_28070
_image removed because it was ugly_
The y-axis shows the different heights (all heights are occupied!), x shows the percentage - exactly as in the link about the hypsographic curve I gave above.

A couple of things can be seen:
- oceans are kind of shallow in comparison to Earth
- the general shape of the curve seems ok (even though the land is a bit too high on average)
- there are some _very_ high places in that world

It would probably be interesting to take a look at this plot right as the heightmap comes out of Platec. Aside from that I hope to finally find good values for [PR 188](https://github.com/Mindwerks/worldengine/pull/188/files#diff-80eed145c1070172ffba1f642a13b731R108) - if somebody can give hints, please do so.

**EDIT:** This also shows that we shouldn't just "fill up" the ocean to an arbitrary value after Platec generated the world making use of a very specific ocean-level. If the ocean_level variable is to be repaired, it should tie into Platecs generation.

**EDIT2:**
A slightly improved plot - there is a vertical line every ten percent now and a horizontal line for the different height thresholds (_sea_, _plains_, _hills_):
hypsographic_28070

Here what it would look like after #188, the thresholds from bottom to top being _sea_, _plains_, _hills_, _low mountains_, _medium mountains_, _high mountains_:
hypsographic_28070

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,3 +5,4 @@ pypng==0.0.18
 PyPlatec==1.4.0
 protobuf==3.0.0a3
 six==1.10.0
+matplotlib
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,8 @@ config = {
         'console_scripts': ['worldengine=worldengine.cli.main:main'],
     },
     'install_requires': ['PyPlatec==1.4.0', 'pypng>=0.0.18', 'numpy>=1.9.2, <= 1.10.0.post2',
-                         'argparse==1.2.1', 'noise==1.2.2', 'protobuf==3.0.0a3'],
+                         'argparse==1.2.1', 'noise==1.2.2', 'protobuf==3.0.0a3',
+                         'matplotlib'],
     'license': 'MIT License'
 }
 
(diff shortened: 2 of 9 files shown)


ftomassetti:
About the fact that ocean are shallow: we can easily rescale that

ftomassetti:
I think this plot could be an interesting addition for the more technical users

tcld:
I changed the code to make use of matplotlib. It is only one commit so far, so I could go back, if necessary.
hypsographic_28070

PR #205, opened 2015-11-18
Biomes manual

esampson:
Created manual page to hopefully explain how --temps, --humids, -go, and -gv work.

I tried to be clear and concise while supplying full details, tasks which are often at odds with one another. Please let me know if things are unclear, redundant, overly verbose, etc.

### DIFF
--- a/manual/biomes.rst
+++ b/manual/biomes.rst
@@ -62,3 +62,5 @@ Each point on this plot is a single point on the landmass of the planet. The Tem
 
 .. image:: https://raw.githubusercontent.com/Mindwerks/worldengine-data/master/images/examples/scatter_plot-labelled.png
    :align: center
+
+Cells in grey are ones which do not technically occur in the standard Holdridge life zone model and as the example above shows there are certain areas of terrain in the example that are both **Polar** and **Semiarid**. Since there are not too many of them we will not concern ourselves as Worldengine will simply treat them as **Polar** and **Arid** (i.e. classify them as **Polar Ice**).
--- a/manual/biomes.rst
+++ b/manual/biomes.rst
@@ -12,11 +12,18 @@ Worldengine implements a version of the `Holdridge life zones <https://en.wikipe
 
 .. image:: https://raw.githubusercontent.com/Mindwerks/worldengine-data/master/docs/Holdridge.png
    :align: center
-   :width: 80%
+   :width: 95%
 
 As an example, an area of the planet that is **Boreal** in temperature and **Semiarid** in humidity will be classified as **Boreal Wet Forest**. 
 
-While many people regard the `Köppen climate classification <https://en.wikipedia.org/wiki/K%C3%B6ppen_climate_classification>`_ system as superior it requires the ability to predict seasonal weather. This would require a highly complicated weather simulation system which Worldengine does not currently have. As a result all Temperature and Humidity values are given in terms of yearly averages.
+While many people regard the |Koppen|_  system as superior it requires the ability to predict seasonal weather. This would require a highly complicated weather simulation system which Worldengine does not currently have. As a result all Temperature and Humidity values are given in terms of yearly averages.
+
+
+.. |Koppen| replace:: K |o| ppen climate classification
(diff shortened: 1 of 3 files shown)


tcld:
I don't know rst-syntax, so I can't claim to have checked it. Aside from that this is an impressive piece of work and a great (and very much appreciated) read. :) Since I was the one complaining about lacking documentation regarding your newly introduced commandline parameters, I sure want to thank you for taking the time to write all of this down as clearly as you did.

I have nothing else to add, this seems flawless. :)

esampson:
I noticed a minor flaw in the output so I corrected it.

I didn't think about the fact that people might not be able to see the output.

http://worldengine-ecs.readthedocs.org/en/latest/index.html

manual/biomes.rst
```diff
@@ -0,0 +1,79 @@
...
+
+Each point on this plot is a single point on the landmass of the planet. The Temperature runs across the bottom, ranging from **Polar** to **Tropical** while the Humidity runs along the side, ranging from **Superarid** to **Superhumid**. Lines are drawn to show the dividing point between various Temperature and Humidity ranges. While the current routine does not label the rows and columns the following image is provided to help understand how they correspond to the chart:
+
+.. image:: https://raw.githubusercontent.com/Mindwerks/worldengine-data/master/images/examples/scatter_plot-labelled.png
+   :align: center
+
+Cells in grey are ones which do not technically occur in the standard Holdridge life zone model and as the example above shows there are certain areas of terrain in the example that are both **Polar** and **Semiarid**. Since there are not too many of them we will not concern ourselves as Worldengine will simply treat them as **Polar** and **Arid** (i.e. classify them as **Polar Ice**).
+
+Biome.png
+---------
```
tcld: Maybe you just want to name it "Biomes"? Since it's a headline that might look nicer.
Is the image itself generated from `docs/Biomes.html`?

manual/biomes.rst
```diff
@@ -0,0 +1,79 @@
...
+
+As an example, an area of the planet that is **Boreal** in temperature and **Semiarid** in humidity will be classified as **Boreal Wet Forest**. 
+
+While many people regard the |Koppen|_  system as superior it requires the ability to predict seasonal weather. This would require a highly complicated weather simulation system which Worldengine does not currently have. As a result all Temperature and Humidity values are given in terms of yearly averages.
+
+
+.. |Koppen| replace:: K |o| ppen climate classification
+.. _Koppen: https://en.wikipedia.org/wiki/K%C3%B6ppen_climate_classification
+
+.. |o| unicode:: 0xf6
```
tcld: For code-readability you could add the comment [from here](http://docutils.sourceforge.net/docutils/parsers/rst/include/isolat1.txt):
`.. |ouml|   unicode:: U+000F6 .. LATIN SMALL LETTER O WITH DIAERESIS`,
i.e. append something similar to this
` .. LATIN SMALL LETTER O WITH DIAERESIS`

You might also be able to just store the file in Unicode-format and write the ö itself. I don't know what this is interpreted by at the end, but I assume proper Unicode might be fine. Although there is a danger of somebody wanting to edit this file one day and messing up the encoding again.

ftomassetti:
Looks good to me but I will wait to merge this because there are a couple of comments from @tcld

manual/biomes.rst
esampson: Yes, the image is just a screenshot of biomes.html. Unfortunately RST doesn't appear to support coloring the backgrounds of table cells so I couldn't just create the table in the document.

Biomes.png isn't a good title, since the output might not be a png, but just Biomes is a little vague. I think Biome Images would be a better title.

manual/biomes.rst
esampson: I tried using unicode initially and somewhere along the line the character was getting nuked, so I think this is a more reliable method. I'll add the comment to make it clearer, however.

tcld:
My comments are addressed and I am very much satisfied now.^^

ftomassetti:
Merging then :)

PR #207, opened 2015-11-19
Reorganize World

ftomassetti:
This is an initial step to reorganize the World class. Other refactoring will follow but I thought it would be better to start discussing and merging this one.

Now most of data is grouped either under generation_params or in layers.

A Layer represent a piece of information available for each cell. Each layer has a matrix of data, some of them have also thresholds while others have quantiles.

The idea is to treat layers in the more homogeneous way possible.


tcld:
Oh. My. God. :D I would vote for merging everything else first so I don't have to fix all of my PRs.^^
Aside from that this is certainly a good idea; the tests were cancelled for some reason, though.

To make sure that there are no bugs left: How about you add another test that basically does what `data_generator.py` does (i.e. generates a world exactly like the one in there)? Then you could compare the "new" world with the "old" one, and if everything checks out, the new code should be fine. Right?

worldengine/draw.py
```diff
@@ -228,25 +230,26 @@ def average_colors(c1, c2):
...
+    mask = numpy.ma.array(e, mask=numpy.logical_not(world.layers['ocean'].data))  # only ocean
     min_elev_sea = mask.min()
     max_elev_sea = mask.max()
     elev_delta_sea = max_elev_sea - min_elev_sea
 
     c = numpy.empty(e.shape, dtype=numpy.float)
-    c[numpy.invert(world.ocean)] = (e[numpy.invert(world.ocean)] - min_elev_land) * 127 / elev_delta_land + 128
-    c[world.ocean] = (e[world.ocean] - min_elev_sea) * 127 / elev_delta_sea
+    c[numpy.invert(world.layers['ocean'].data)] = (e[numpy.invert(world.layers['ocean'].data)] - min_elev_land) * 127 / elev_delta_land + 128
+    c[world.layers['ocean'].data] = (e[world.layers['ocean'].data] - min_elev_sea) * 127 / elev_delta_sea
```
tcld: At this point I would suggest introducing a reference to the ocean at the very top of the function; it will make thinks more readable.

worldengine/draw.py
```diff
@@ -693,7 +696,7 @@ def draw_scatter_plot(world, size, target):
         for x in range(world.width):
             if world.is_land((x, y)):
                 t = world.temperature_at((x, y))
-                p = world.humidity['data'][y, x]
+                p = world.humidity_at((x, y))
```
tcld: This is something I keep thinking about. Both notations show up at various places and while the latter is a littler prettier, is there an actual chance of this becoming important? There are millions over millions of function-calls triggered by the "pretty kind of notation" ( and it's O(n²) ); especially calling things like `is_land()` contribute a non-insignificant amount of time to the overall generation time. I'm not sure if that is truly worth it.

ftomassetti: for me it is not about being pretty or not but it is about maintanability through encapsulation: if the latter one was used this patch would be like ten times smaller because we could have changed the way we keep data without affecting the whole codebase. I agree that it could cause a performance hit but in the cases where it is strictly needed we can access directly the array. In many cases we could use numpy functions instead of accessing single values.

tcld: But is there any chance at all that `humidity_at((x, y))` will ever not return the same value as `humidity['data'][y, x]`? For something like `is_land()` that might be the case, since the decision could eventually have some layers - but for most variables I don't see that being the case.

Calls to numpy-functions usually undermine the concept of maintainability anyway, the arrays are directly used instead of via a call to a centralized function. So maintainability in the face of renaming variables or restructuring the code isn't given either - with none of the notations.

In short: I understand the reasoning, but I don't see a realistic case where the methodology would ever pay off. And until then it is going to eat performance.

worldengine/hdf5_serialization.py
```diff
@@ -16,87 +18,88 @@ def save_world_to_hdf5(world, filename):
 
     elevation_grp = f.create_group("elevation")
     elevation_ths_grp = elevation_grp.create_group("thresholds")
-    elevation_ths_grp["sea"] = world.elevation['thresholds'][0][1]
-    elevation_ths_grp["plain"] = world.elevation['thresholds'][1][1]
-    elevation_ths_grp["hill"] = world.elevation['thresholds'][2][1]
+    elevation_ths_grp["sea"] = world.layers['elevation'].thresholds[0][1]
+    elevation_ths_grp["plain"] = world.layers['elevation'].thresholds[1][1]
+    elevation_ths_grp["hill"] = world.layers['elevation'].thresholds[2][1]
```
tcld: Should these maybe saved with a for-loop? In case more are ever added.^^ (applies to the protobuf-save, too)

worldengine/draw.py
ftomassetti:
```
But is there any chance at all that humidity_at((x, y)) will ever not return the same value as humidity['data'][y, x]?
```
Yes: this is what happens when we change how we store data as we are doing in this PR. When we access humidity['data'] from outside world we depends on the internal structure of World, not on its interface and that means there is no encapsulation: if we change how World is organized every usage of World is affected, which is very bad from the maintanability point of view.
If instead we use proper function we can pass by having internally world.humidity to having world.layers['humidity'] without having to change any usage.

About using numpy functions we could do that through interfaces, passing lambdas or functions or exposing numpy functions (as we do with min and max which are exposed by the Layer class).

I would say we should strive for maintanability and only where there is a realy performance hit we could go for the "hack" where the hack is accessing directly internal world variables.

worldengine/model/world.py
```diff
@@ -1,43 +1,113 @@
...
-    CoolTemperateMoistForest, CoolTemperateRainForest, CoolTemperateSteppe, CoolTemperateWetForest,\
-    Ice, PolarDesert, SubpolarDryTundra, SubpolarMoistTundra, SubpolarRainTundra, \
-    SubpolarWetTundra, SubtropicalDesert, SubtropicalDesertScrub, SubtropicalDryForest, \
-    SubtropicalMoistForest, SubtropicalRainForest, SubtropicalThornWoodland, SubtropicalWetForest, \
-    WarmTemperateDesert, WarmTemperateDesertScrub, WarmTemperateDryForest, \
-    WarmTemperateMoistForest, WarmTemperateRainForest, WarmTemperateThornScrub, \
-    WarmTemperateWetForest, TropicalDesert, TropicalDesertScrub, TropicalDryForest, \
-    TropicalMoistForest, TropicalRainForest, TropicalThornWoodland, TropicalWetForest, \
-    TropicalVeryDryForest, biome_index_to_name, biome_name_to_index
+from worldengine.biome import *
```
tcld: I normally don't like the *-import, but this is a total exception. :D

worldengine/model/world.py
```diff
@@ -132,11 +202,9 @@ def _from_protobuf_quantiles(p_quantiles):
 
     @staticmethod
     def _from_protobuf_matrix_with_quantiles(p_matrix):
```
tcld: I think I mentioned this recently, but how about putting all the Protobuf-stuff into its own file while sticking as closely as possible to how things are done in the HDF5-file? The saving of both formats is extremely similar, and all the Protobuf-code is cluttering up this file a bit.

ftomassetti: I agree with that but I would not do that as part of this PR because this is already big enough :)

ftomassetti:
Given this one would have quite an impact it would nice to have the opinion of @psi29a before merging it

worldengine/model/world.py
tcld: You are changing almost every line anyway, the PR wouldn't really get much bigger, I think.^^

tcld:
I looked over it once, comments above.

I guess this should be merged as soon as possible, since it will break every piece of code that isn't already merged. (I would prefer if #188 and #203 could be merged first, so I don't have to invest more work... but overall it might even be easier if they were merged after this PR.)

I also want to mention all the used "thresholds" again - it would be really nice if they could just be stored via for-loop. It would make it slightly easier to modify/add thresholds.

ftomassetti:
About:
```
To make sure that there are no bugs left: How about you add another test that basically does what data_generator.py does (i.e. generates a world exactly like the one in there)? 
```
I think that would not work because platec could give different results on different platforms (for example because of rounding or different optimization with floating numbers)

worldengine/draw.py
ftomassetti: done

tcld:
Maybe if you started from a plate-only world?

psi29a:
Great googly moogly... I'll give this an eye-ball around lunch today.

ftomassetti:
GitHub did not told me you edited the comment... I will reach you on GTalk. The main idea is to put some order in the World class:
* I created a class named GenerationParameters: ideally from an instance of this class we should be able to regenerate a world
* I collected all the different kinds of data in a map of layers. So you have layers['elevation'], layers['ocean'] and so on

Each layer is an instance of Layer, LayerWithThresholds or LayerWithQuantiles. All of these classes have a field named data which is a numpy array. Then they could have also thresholds or quantiles :)

The idea is to make layers to be more homogeneous: before  world.elevation['data'] and world.ocean were numpy arrays. Now you would access both as world.layers['elevation'].data and world.layers['ocean'].data.

When we add a new piece of information (e.g., wind) we just add a new layer.

Aside from that I tried to use encapsulation when possible, so using method such as world.is_ocean(pos) instead of accessing directly variables (e.g., world.ocean[y, x] or world.elevation['data'][y, x]).  This should make the code more explicit and more robust to changes. I left direct access to internal variables where it made sense for performance reasons.

tcld:
Since you mention performance, I wrote a [little script](https://gist.github.com/tcld/4f6e1ddcc5ed3b726e67) yesterday. It accesses an array 50 million times via different means, that's around the order of accesses WorldEngine performs for a medium-sized world.
Accessing an array via a separate function is not that bad compared to the overall generation time (it might amount to ~10%).
```
System: linux2
Python: v2.7.6
 Numpy: v1.8.2

                   Empty function (50.0M calls):  5.869s
         Direct access via [y, x] (50.0M calls):  8.574s
                  Function access (50.0M calls): 15.836s
Function access w/ unpacked tuple (50.0M calls): 14.737s
         Direct access via [y][x] (50.0M calls): 15.560s

 Python for-loop (10 iterations):  0.885s
  Numpy for-loop (10 iterations):  5.684s
Numpy while-loop (10 iterations):  4.593s
```

```
System: linux
Python: v3.4.3
 Numpy: v1.9.2

                   Empty function (50.0M calls):  5.736s
         Direct access via [y, x] (50.0M calls):  6.701s
                  Function access (50.0M calls): 11.592s
Function access w/ unpacked tuple (50.0M calls): 11.420s
         Direct access via [y][x] (50.0M calls): 12.282s

 Python for-loop (10 iterations):  0.995s
  Numpy for-loop (10 iterations):  4.619s
Numpy while-loop (10 iterations):  3.939s
```

An additional suggestion, to keep things clear and clean, would be to add new functions to `World` to request a view on the whole array at once, similar to one of these:
```python
def temperature(self, x1=None,y1=None,x2=None,y2=None):
    return self.layers['temperature'][y1:y2, x1:x2]

def temperature(self):
    return self.layers['temperature']

def layer(self, layer):
    return self.layers[layer]
```
So when somebody wants to do fancy stuff with numpy, e.g. `output = w.layers['temperature'] / w.layers['humidity']`, you can still hide away the actual access of the array and keep control within `ẁorld.py` 100% of the time: `output = w.temperature() / w.humidity()`

ftomassetti:
We need to merge this one :)

psi29a:
No real comments, and looks good to me.

ftomassetti:
Great, thanks!

Issue #177, opened 2015-11-03
Ancient Map Upgrades


esampson:
I'm starting some experiments with upgrading the ancient map routines. Just as a proof of concept I modified the mountains.
![ancient_map_ancient_test-old](https://cloud.githubusercontent.com/assets/8743461/10902083/c2d611bc-81ad-11e5-99a9-6449091c1dc7.png)
What I'm doing is basically just a replacement of the draw_mountain routine. Instead of the existing routine I am calling a new routine called '_stamp' that is fed a bitmap. The current setup is extremely rudimentary. I have 4 different PNG files that are loaded and I randomly feed one of them to the stamp routine.

So we could probably replace all of the individual _draw_(something) routines with the _stamp routine when an object is being draw, which would probably be good. I've also got an idea for handling areas that are textures such as ice and steppes so we could replace those remaining routines with this second routine.

One of the questions I have is whether we want to have the final system load bitmap files or whether we want to embed the data in the program itself using techniques such as PIL's image.fromstring() function. Loading individual PNG bitmaps would certainly be easier, program-wise, but it would also probably result in a lot of files being added as well.

ftomassetti:
Not bad, not bad at all...

My first comment is that the mountain chosen should not be completely random but instead depend on the elevation (higher elevation, higher image). Perhaps you could dynamically rescale the images also.

I would be perfectly fine to add a lot of files to the project but on this it is better to hear the opinion of @psi29a who did a lot of work to clean the repository after I made it grow to a few hundreds megabytes...

esampson:
Actually if you look carefully you can see that it does rescale the images. There are some smaller mountains.

As for doing things like using certain mountains when the elevation is higher, that would be extremely easy. All we do is go into the routine that is placing all the mountains and rather than simply choosing one at random we first check elevation. If it is over a certain amount we randomly choose from a set of images for tall mountains, if it is between a certain range we choose from a set of regular mountains and if it is below that range we choose from a set of hills.

Since we don't have to write a new routine for each new type of image it is about 50 times easier to add new images then it was before. Additionally, if we continue to load PNG images for our icons that should put us in a very good position for letting other people 'reskin' the ancient map routine with their own sets of icons. At its simplest we just have an extra command line option for drawing ancient maps that specifies an alternate directory for the icons.

ftomassetti:
"Actually if you look carefully you can see that it does rescale the images. There are some smaller mountains."

I did not know if they were just different images :)

[...]

PR #208, opened 2015-11-20
PDF creation from RTD

esampson:
It turns out that while the images show up if you are viewing the manual online through RTD they don't show up properly in the downloaded HTMLzip or PDFs unless the images are moved into a subdirectory of the directory containing the RST file.

Additionally I had to add some specific commands when rendering the PDF through LaTex to make the images come out better.


psi29a:
I'm going to veto this, no binary data in worldengine please.

What you can do is use the URL to worldengine-data instead.

If that isn't solvable, then we need to move our docs to worldengine-data.

esampson:
We can move the docs to worldengine-data. That should be very easy. As I understand it all we need to do is copy the files and then change the entry in ReadTheDocs to point to the different git distribution.

There is one issue, however, that is indirectly connected to this. The work that I'm doing with an alternative system for drawing ancient maps (https://github.com/Mindwerks/worldengine/issues/177) requires at least one PNG file (or some other form of non-readable file. I could use an asciii encode of the PNG, but what would be the point to that?).

We could start a third git repository for alternate styles but I think we should include at least a single default style so that the routine works 'out of the box'. There are other options. The first I can think of is that I believe I can reference the file remotely (through http), but that would mean that the computer has to have an internet connection to use the routine, which I don't like. The other possibility would be to make the new routine a plugin and give it and other plugins their own git repository.

esampson:
Oh. I should add that my first version did use the URL to worldengine-data. While it produced good results when looking at the ReadTheDocs website the images were not included in the HTMLzip or PDF files.

esampson:
Just a quick note to confirm, you can simply redirect the document project to an alternate git repository.

PR #211, opened 2015-11-28
Update __init__.py - resolve Windows error 32

DevinRB:
When running export in Windows, errors out: "WindowsError: [Error 32] The process cannot access the file because it is in use by another process: path_to_temp_file"

Flushing (new line 132) and adding a call to close with the file handle (new line 134) resolves the error. Just flushing, or just closing, does not.

Thanks!


psi29a:
Super find! I'll merge it.

It worked on my release, but it could be python distribution dependent (how that python is installed on the system). Might I ask what your setup was? It might be useful for in the changelog in the future?

Issue #209, opened 2015-11-23
Add property decorators in World


ftomassetti:
For example:

```python
@property
def elevation:
    return self.layers['elevation'].data
```

tcld:
It might not hurt if you add this to #207, if you can spare the time. By "add" I mean "create all necessary functions and then replace every use of any layer throughout the code with a call to the corresponding property".

Also, do you need to explicitly create a setter, too?
```
@property
def elevation(self):
    return self.layers['elevation'].data

@elevation.setter
def elevation(self, value):
    self.layers['elevation'].data = value
```

ftomassetti:
Me and Bret discussed it and we would prefer to go by steps: so we first merge #207 and then introduce property decorators. I am not sure about the setters perhaps it could make sense given we had all those set_ocean, set_temperature methods to replace

psi29a:
Thanks for this @tcld it is good plan. :)

Himura2la:
I'm going to convert to properties all functions that starts with `set_`. I'll also try to fins their `get_` brothers

PR #214, opened 2015-12-04
Issue 209: Add layers accessors

ftomassetti:
See issue #209

Note that the accessors return the whole layer, then the data field has to be accessed explicitely. Also theresholds and quantiles can be accessed from the layer.

### DIFF
--- a/worldengine/hdf5_serialization.py
+++ b/worldengine/hdf5_serialization.py
@@ -25,61 +25,61 @@ def save_world_to_hdf5(world, filename):
     elevation_data.write_direct(world.elevation.data)
 
     plates_data = f.create_dataset("plates", (world.height, world.width), dtype=numpy.uint16)
-    plates_data.write_direct(world.layers['plates'].data)
+    plates_data.write_direct(world.plates.data)
 
     ocean_data = f.create_dataset("ocean", (world.height, world.width), dtype=numpy.bool)
-    ocean_data.write_direct(world.layers['ocean'].data)
+    ocean_data.write_direct(world.ocean.data)
 
     sea_depth_data = f.create_dataset("sea_depth", (world.height, world.width), dtype=numpy.float)
-    sea_depth_data.write_direct(world.layers['sea_depth'].data)
+    sea_depth_data.write_direct(world.sea_depth.data)
 
     if world.has_biome():
         biome_data = f.create_dataset("biome", (world.height, world.width), dtype=numpy.uint16)
         for y in range(world.height):
             for x in range(world.width):
-                biome_data[y, x] = biome_name_to_index(world.layers['biome'].data[y][x])
+                biome_data[y, x] = biome_name_to_index(world.biome.data[y][x])
 
     if world.has_humidity():
         humidity_grp = f.create_group("humidity")
         humidity_quantiles_grp = humidity_grp.create_group("quantiles")
-        for k in world.layers['humidity'].quantiles.keys():
-            humidity_quantiles_grp[k] = world.layers['humidity'].quantiles[k]
+        for k in world.humidity.quantiles.keys():
+            humidity_quantiles_grp[k] = world.humidity.quantiles[k]
         humidity_data = humidity_grp.create_dataset("data", (world.height, world.width), dtype=numpy.float)
-        humidity_data.write_direct(world.layers['humidity'].data)
+        humidity_data.write_direct(world.humidity.data)
 
     if world.has_irrigation():
         irrigation_data = f.create_dataset("irrigation", (world.height, world.width), dtype=numpy.float)
-        irrigation_data.write_direct(world.layers['irrigation'].data)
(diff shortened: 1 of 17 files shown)


tcld:
This looks *so much* nicer and cleaner than before! I have not checked every line yet.

One thing I mentioned before: Why not store the thresholds in loops? (I'll add an example comment for that.)

worldengine/hdf5_serialization.py
```diff
@@ -18,68 +18,68 @@ def save_world_to_hdf5(world, filename):
...
-        irrigation_data.write_direct(world.layers['irrigation'].data)
+        irrigation_data.write_direct(world.irrigation.data)
 
     if world.has_permeability():
         permeability_grp = f.create_group("permeability")
         permeability_ths_grp = permeability_grp.create_group("thresholds")
-        permeability_ths_grp['low'] = world.layers['permeability'].thresholds[0][1]
-        permeability_ths_grp['med'] = world.layers['permeability'].thresholds[1][1]
+        permeability_ths_grp['low'] = world.permeability.thresholds[0][1]
+        permeability_ths_grp['med'] = world.permeability.thresholds[1][1]
```
tcld: I would much prefer this when it comes to storing thresholds:
```python
for pair in world.permeability.thresholds:
    permeability_ths_grp[pair[0]] = pair[1]
```

As far as I understand the item at [x][0] is a naming string anyway. So far I haven't seen those strings used anywhere - so they could be renamed to 'low', 'med' etc., if necessary. Then the stored file would be the same as before, but the whole system could be expanded/changed more easily.

worldengine/model/world.py
```diff
@@ -852,7 +851,67 @@ def is_chaparral(self, pos):
...
-        return self.layers['plates'].data.max() + 1
+        return self.plates.data.max() + 1
+
+    #
+    # Accessors
+    #
+
+    @property
+    def elevation(self):
+        return self.layers['elevation']
```
tcld: I have not done much with `@property` myself so far, so maybe this is a stupid question: Is there no need for an explicit `set()`-function? Will Python create one on its own? [The properties are currently set via their own function](https://github.com/Mindwerks/worldengine/blob/issue209/worldengine/simulations/irrigation.py#L11), so the whole "property"-idea isn't fully made use of in my opinion.
```python
@property
def elevation(self):
    return self.layers['elevation'].data

@elevation.setter
def elevation(self, value):
    self.layers['elevation'].data = value
```
That second part is missing.

ftomassetti: Well, some setters here take one or two values (data and theresholds or data and quantiles) and then instantiate a class (Layer, LayerWithQuantiles or LayerWithThresholds). So to have the setter we should demand users to instantiate that class and assign it to the properties. I would rather hide that in a setter. I think without the property setter we just miss the "syntax sugar". It would be cool to have it but it is not strictly needed.

tcld: You should be able to do whatever you want in the `.setter`-function, it is not limited to the example I gave.

ftomassetti: Yes, sure but if the property reader return a layer the property setter should accept a layer (otherwise it would be really weird from the user point of view) and so the user would have to instantiate those Layer classes.

tcld: I don't think that would be weird. Maybe an overload would be possible, though (I don't know if that actually works).

ftomassetti: currently many setters take 2 values and we could not replicate that for property setters. For the others, if we implemented propery setters accepting a numpy matrix it would like this:

```
world.ocean = data
data = world.ocean.data
```

So things like:
```
world.ocean = world.ocean
```

Would not work and actually mess up things, you should instead use

```
world.ocean = world.ocean.data
```

So I would stay away from property setters and use explicit setters for now

tcld: That last problem can be handled easily.
```python
@elevation.setter
def elevation(self, value):
    if type(value) is not LayerWithThresholds:
        value = LayerWithThresholds(value, None)
    self.layers['elevation'].data = value
```

I am not sure about multiple parameters, though. Doesn't `world.elevation = (data, thresholds)` work?

ftomassetti: Yes, it is possible to have different behavior looking at the type of the value and yes, we could pass a tutple to the property setter but I am very convinced this would cause confusion. I would strongly prefer using explicit setters as we have now. Anyway setters are used very rarely, while property getters are used much more frequently, so they are much more important

tcld: I'm not sure if a line like `world.elevation = (data, thresholds)` is confusing at all - look at it without any context, as if you didn't know anything about WorldEngine. (I updated the code-example above so it is actually correct.)
You are right in that the setters aren't used too much, I just think that if you want to use properties you may want to go all the way. Having a `set_elevation()` and not a `get_elevation()` is pretty ugly in and of itself!

PR #215, opened 2015-12-05
Move option for the number of plates among generation options

ftomassetti:


tcld:
Makes sense, looks good.^^

PR #217, opened 2015-12-05
Winds: initial winds implementation

ftomassetti:
This is based on this: http://www.dungeonleague.com/2010/03/28/wind-direction/

This PR is based on top of #214

Example of generated winds:
seed_27421_wind

### DIFF
--- a/worldengine/simulations/wind.py
+++ b/worldengine/simulations/wind.py
@@ -1,0 +1,21 @@
+from worldengine.simulations.basic import find_threshold_f
+import numpy
+
+
+class WindSimulation(object):
+
+    @staticmethod
+    def is_applicable(world):
+        return not world.has_wind()
+
+    def execute(self, world, seed):
+        assert seed is not None
+        data = self._wind(world, 1.5)
+        world.set_wind(data)
+
+    @staticmethod
+    def _wind(world, distorsion_factor):
+        print('WIIIIND')
+        print('WIIIIND')
+        print('WIIIIND')
+        return None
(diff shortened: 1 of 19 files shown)


tcld:
This is based on #214 , right? I will wait until that one is polished and merged before looking at the code in here then.

The map looks interesting. :)

worldengine/simulations/wind.py
```diff
@@ -0,0 +1,102 @@
...
+    def is_applicable(world):
+        return not world.has_wind()
+
+    def execute(self, world, seed):
+        assert seed is not None
+        direction = self._wind_direction(world, 0.5, seed)
+        world.set_wind_direction(direction)
+
+    @staticmethod
+    def _wind_direction(world, distorsion_factor, seed):
```
tcld: I think this should be called `_calculate()` - that's how it is done in all the other modules, right?

worldengine/simulations/wind.py
```diff
@@ -0,0 +1,102 @@
...
+        SOUTH = 0.5
+        WEST = 0.75
+
+        def _set_line_to_gradient(data, y, start_y, end_y, start_value, end_value):
+            delta = float(end_y - start_y)
+            start_affinity = float(end_y - y) / delta
+            end_affinity = 1.0 - start_affinity
+            value = start_value * start_affinity + end_value * end_affinity
+            for x in range(world.width):
+                data[y, x] = value
```
tcld: This can be done all at once. `data[y] = value` should do the trick.

worldengine/simulations/wind.py
ftomassetti: makes sense (and yes it is copied :D)

worldengine/simulations/wind.py
ftomassetti: great, I suspected it but I really should learn numpy

PR #223, opened 2016-03-28
move over to using trusty

psi29a:


psi29a:
This switches us away from docker to using a full VM using trusty. With all the things we do with WE, it is easier for us to do our job with a full VM.

psi29a:
Yay, python3.4 tests are testing... 

but this came up, someone fix it! :P
```
  File "/home/travis/build/Mindwerks/worldengine/tests/serialization_test.py", line 9, in <module>
    from sets import Set
ImportError: No module named 'sets'
```

psi29a:
Damn you pypy, so close!

ImportError: No module named google

psi29a:
Yay, tests are running!

psi29a:
They fail on pypy of course, this needs fixing.

h5py doesn't run on pypy.

psi29a:
@ftomassetti is this OK?

psi29a:
OK, merging in.

This means that any future updates have to pass python27, python34, pyflakes and manifest. This will helps keep WE stable.

PR #242, opened 2017-09-16
Make sea_depth much faster with dynamic programming

MM1nd:
Hi Folks,

in 2016 I played around with worldengine and found it to be slow to the point of not being fun to use. Maybe my machine is to blame, I don't know. I found that a lot of speed could be gained, not by dreaded "premature optimisation" but by simply thinking more clearly about what we are trying to accomplish and using numpy better. Unfortunately I never got around to pushing any of that work upstream. 

One of the things I found while profiling was that worldengine will spend ~15% of total runtme in is_land on my machine for a 128*128 map with otherwise default parameters. I don't think that is acceptable and I think we can do much better. 

Unfortunately I was struggling with git while I did this so the pull request contains my additions to .gitignore and a series of unneccessary commits just to get back to status quo of readme.md. Other than that, the changes are straightforward. 

All tests are green after the change. It's functionally equivalent to the old slow implementation.

If you find this helpful I have made many more performance improvements locally which I'll gladly share over time.

Let me know what you think.

Cheers 

Alex


MM1nd:
P. S. I'll make better use of branches next time. I was confused about how branches work when I did this.

psi29a:
Reviewing... but looks promising. :)

psi29a:
Well damn, not only is there code... there's test code and I don't see any regressions. Going to do a little testing on my side but looks good to me.

@ftomassetti what do you think?

tests/generation_test.py
```diff
@@ -37,6 +42,55 @@ def test_center_land(self):
...
+                                0.9, 0.7, 0.5, 0.3, 0.0, -1.0, 0.0, 0.3, 0.5, 0.7, 0.9, \
+                                0.9, 0.7, 0.5, 0.3, 0.0, 0.0, 0.0, 0.3, 0.5, 0.7, 0.9, \
+                                0.9, 0.7, 0.5, 0.3, 0.3, 0.3, 0.3, 0.3, 0.5, 0.7, 0.9, \
+                                0.9, 0.7, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.5, 0.7, 0.9, \
+                                0.9, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.7, 0.9, \
+                                0.9, 0.9, 0.9, 0.9, 0.9, 0.9, 0.9, 0.9, 0.9, 0.9, 0.9])
+
+        desired_result = desired_result.reshape([extent,extent])
+
+        #this part is verbatim from the function. It's not part of the test
```
ftomassetti: nitpick: I would put a space after the sharp symbol

worldengine/generation.py
```diff
@@ -141,18 +141,53 @@ def harmonize_ocean(ocean, elevation, ocean_level):
...
+
+    next_land = next_land_dynamic(world.layers['ocean'].data)
+
     sea_depth = sea_level - world.layers['elevation'].data
     for y in range(world.height):
         for x in range(world.width):
-            if world.tiles_around((x, y), radius=1, predicate=world.is_land):
+            dist_to_next_land = next_land[y,x]
+
+            if dist_to_next_land == 1:
```
ftomassetti: Maybe we can now use an array containing the different factors and then just do:
```
sea_depth[y, x] *= factor[dist_to_next_land]
```

MM1nd: Of course, Sometimes you don't see the simplest things. In my defense it would have been a bit harder to see then, that the code actually behaves the same.

MM1nd:
I pushed the suggested changes. Thank's for the feedback.

ftomassetti:
For me is ok to merge it, leaving it to @psi29a

MM1nd:
Yeah, travis failed with a type problem not present on my machine. Investigating.

MM1nd:
Fixed the type problem. Should be ready for merge now.

psi29a:
LGTM, merge merge merge!

psi29a:
One last thing, can you add an entry to:

https://github.com/Mindwerks/worldengine/blob/master/CHANGELOG.md

that we know what you did instead of having to hunt it back down later before releasing? :)

After that, it's merge merge merge.

MM1nd:
Done.

psi29a:
Merged, thank you! :)

PR #245, opened 2017-09-30
Faster watermap

MM1nd:
Hi,

this includes #243 and makes use of it. See discussion there. If you are good with this, #243 can be discarded.

Timings with profiler turned on:

Parameters:
-x 1024 -y  512 -v --seed=7577

Current Master:
5m18s, 22,3% in _watermap

This PR:
4m37s 0,5% (!) in _watermap

Times are a bit inflated by disk I/O, my hard drive (still a rotating thing) was doing something else during both tests. However, I think the relevant comparison is legit.

Cheers
Alex


tests/simulation_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+import numpy
+
+from worldengine.simulations.hydrology import WatermapSimulation
+from worldengine.model.world import World, Size, GenerationParameters
+
+class TestSimulation(unittest.TestCase):
+    
+
+    # The hydrology simulation indirectly calls the global rng.
```
ftomassetti: Does some tof this comment belong in the implementation?

tests/simulation_test.py
```diff
@@ -0,0 +1,93 @@
...
+    # In fact it is hard to tell how "correctness" of a Monte Carlo process should
+    # be judged.
+    # Should it ever be decided that a "correct" implementation needs fewer or 
+    # more calls to the rng, relative to the n parameter,
+    # of course compatibility will be broken. 
+    # Implicitly this is tested by the blessed images but it might be useful to
+    # have a finer grained picture.
+
+    def test_watermap_rng_stabilty(self):
+
```
ftomassetti: nitpick: I think we normally do not put a white line after the header of the function

psi29a:
We highly recommend PyCharm, it's free and highly recommended. :)

psi29a:
LGTM, merging.

Thanks! :)

PR #247, opened 2017-10-01
transpose next_land_dynamic for greater speed and clarity

MM1nd:
Here's another one I prepared yesterday.

This uses the fact that only very few coodinates are actually `dist` away from land. It then iterates over these coordinates and looks if the distance of their neighbours happens to be unknown. Only then an update is made. This is actually the more logical thing to do because it checks for `dist` directly at the beginning of the loop iterating over `dist` and it checks `next_land[ny,nx]` in the inner loop and makes a change to the same coordinate. 

The original implementation had this backwards. It checked for unknown distances in the outer loop and for the known distance in the inner loop which makes for a lot more checks because of how the numbers work out.

I think that change in ordering is where the main speedup comes from. Doing the more logical thing also saves two comparisons in the inner loop and is likely to be a tiny bit more cache friendly, but I don't think these things matter too much. Using numpy here makes the code look a bit cleaner but mainly it hides away some of the loops, so I don't think it is a great speedup in and of itself. I didn't measure that in detail though. The code is now reasonably fast and it looks cleaner so that was where I stopped investigating.

Cheers

Alex

### DIFF
--- a/worldengine/generation.py
+++ b/worldengine/generation.py
@@ -158,18 +158,16 @@ def sea_depth(world, sea_level):
         height, width = ocean.shape
 
         for dist in range(max_radius):
-            for y in range(height):
-                for x in range(width):
-                    if next_land[y,x] == -1:
+            indices = numpy.transpose(numpy.where(next_land==dist))
+            for y, x in indices:
+                for dy in range(-1, 2):
+                    ny = y + dy
+                    if 0 <= ny < height:
                         for dx in range(-1, 2):
                             nx = x + dx
                             if 0 <= nx < width:
-                                for dy in range(-1, 2):
-                                    ny = y + dy
-                                    if 0 <= ny < height and (dx != 0 or dy != 0):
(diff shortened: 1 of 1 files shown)


psi29a:
Do we have some before and after numbers here? I love numbers. :)

MM1nd:
Like I said in the other thread, this is the second to last improvement I have prepared in this series, so don't expect wonders. We are clearly reaching a limit here as the remaining slowness increasingly is in foreign code (platec, protobuf, the latter we could maybe use better but *shrug*). I also get the feeling that times being dominated py protobuf makes timings very unstable.

That being said:

Parameters same as always (1024*512 map).

Current master:

2m 30s, ~12% in `next_land_dynamic`

This branch merged into current master:

2m 23s ~2.3% in `next_land_dynamic`` 

But still, it's something and it makes the code more readable. I think it's the latter that counts here.

psi29a:
Thanks! The changes themselves are great, I like that it has become more readable.

PR #248, opened 2017-10-01
Refactor drawing functions

MM1nd:
Hi again,

this is at best the first step on a long journey to make `ancient_map` faster. However, this is huge already, so I'm making it a PR before continuing. Note that while there is one more PR to come for `generate_world` this is not related to those changes I made a year ago. This is new.

It does a lot of things so let's explain:

**Biomes**
- `ancient_map` groups several biomes together. For example `BorealMoistForest`, `BorealWetForest` and `BorealRainForest` are grouped together as `BorealForest`. The information, how this grouping is done was hardcoded in the implementation of some functions in `world.py` e. g. `is_boreal_forest`.
- I argue that this is not a good idea. Because a) I need this information to make ancient_map fast and there is nowhere to read it from other than the code itself and b) it's inconsistent. The place to put all information regarding biomes, and that includes their grouping, should be in `biome.py`
- The changes to `biome.py` make good on the concept established earlier therein to model biomes through the class hierarchy. These changes are straightforward and consistant with how biomes always worked.
- We can provide functions in `world.py` who use this and work exactly like the old `is_*`functions. There is one example left and that is [`is_iceland`](https://github.com/Mindwerks/worldengine/compare/master...MM1nd:refactor_drawing_functions?expand=1#diff-ab53213f333842a1679f6c236eddf92aR722). The other functions of that sort can be written in the same style, at some point during the refactoring I had them all there, but eventually nobody calls them, there is no testing code, and they were terribly slow. So I decided to remove them. I know that might be controversial, but I really think it's for the better. I'm pretty sure they were only added for the one use case that's no longer there, anyway.
- This also means that `world.py` needs to import far fewer things. 
- The end game here is to get rid of `tiles_around` and the like entirely. I can see the beauty of the concept but these functions are excruciatingly slow and they are not easier to use than manipulating the data directly with numpy. No offense, but I think this was overengineering, it did not work out, and it should no longer be there.

**Bugfix**
- `draw_ancient_map` was called with `verbose=False` disregarding the command line. I fixed that in `main.py`

**Quality of Life Improvement**
- I added a function `count_neighbours` to `common.py` that given an array of binary values can count how many neighbours of each coordinate are set to one. It is fast.

**Refactorings**
- There was a lot of code in `drawing_functions.py` that screamed "copy and paste".  The github diff is horrible at showing this to the point of being useless, but there are only three different pixel patters that where used in several functions so I extracted them to be reused (`_draw_forest_pattern1`, `_draw_forest_pattern2`, `_draw_desert_pattern`).
- In the same vein, the shading method for most of the other biomes was the same fundamentally so I factored that out too (`_draw_shaded_pixel`).
- Note that this PR comes with 237 additions and 465 deletions while being functionally equivalent and faster. This is not the result of trying to make the code short at the cost of becoming obscure, I think the readability went up. 

**Speedup**
- Calculation of masks for the individual biomes was extremely slow and used a lot of code. With the new biome groups it is all done generically in `_build_biome_group_masks`. All other mask related code is simply gone. This makes it super easy to add new biome groups.
- The rest of the changes in `drawing_functions.py` make it so that the generic masks are used. These changes are straightforward.
- As for the effect, this brings `test_ancient_map` to unter a minute for the first time on my machine. It takes a full minute off the clock for drawing a 1024*512 map at scaling factor 1 (from ~4 minutes to ~3 minutes)

**Next steps**
- There is another `anti_alias` function here, distinct from the one in `common.py` and it's very slow. With this PR everything is prepared to tackle that issue next.

Cheers
Alex

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 Version 0.20 (UNRELEASED)
 
 * Calculation of sea depth is now faster.
+* Hydrology simulation is now faster.
 
 Version 0.19
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ Version 0.20 (UNRELEASED)
 
 * Calculation of sea depth is now faster.
 * Hydrology simulation is now faster.
+* BiomeGroups are now configurable via the class hierarchy.
+* Ancient map now initialises faster.
 
 Version 0.19
 
--- a/tests/common_test.py
+++ b/tests/common_test.py
@@ -1,6 +1,6 @@
 import unittest
 import numpy
-from worldengine.common import Counter, anti_alias, get_verbose, set_verbose
+from worldengine.common import Counter, anti_alias, get_verbose, set_verbose, _equal
 
 
 class TestCommon(unittest.TestCase):
@@ -37,6 +37,18 @@ class TestCommon(unittest.TestCase):
         original = numpy.array([[0.8]])
         antialiased = anti_alias(original, 10)
         self.assertAlmostEquals(0.8, antialiased[0][0])
+
+    def test_dictionary_equality(self):
+        a = {}
+        b = {}
+
+        a['foo'] = 'bar'
+
+        self.assertFalse(_equal(a,b))
+
+        b['foo'] = 'bar'
+
+        self.assertTrue(_equal(a,b))
 
 if __name__ == '__main__':
     unittest.main()
--- a/tests/simulation_test.py
+++ b/tests/simulation_test.py
@@ -56,6 +56,22 @@ class TestSimulation(unittest.TestCase):
         d = numpy.random.randint(0,100)
         self.assertEqual(d, 59)
 
+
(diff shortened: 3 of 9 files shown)


ftomassetti:
The code definitely looks nicer, if we are sure it also produces the same result... it is quite an achievement!

MM1nd:
"if we are sure it also produces the same result... it is quite an achievement!"

I triple checked at all stages. Not saying its impossible that I missed something, but it seems unlikely. Ideally you make a reference image and see for yourself. The tests are good (currently travis fails for a pyflakes annoyance, I'm about to fix that)

psi29a:
I'm overwhelmed by how professional your posts/PRs are, from documentation to code itself. Great job!

MM1nd:
Well, thank you!

psi29a:
time python worldengine -x 1024 -y 1024
went from
 real	5m43.636s
to
 real	2m5.086s

Great work!

Issue #246, opened 2017-09-30
Consolidated discussion of ideas related to platec


MM1nd:
Hi,

1) In my current local version (#245), platec.step is the biggest time consumer. I suspect that this will get relatively worse when I continue to improve on the other parts. I also suspect it will be much harder to make platec faster because the easy way of "let's move this to numpy" is not available. I haven't looked into the platec code base, maybe there are some low hanging fruits, but eventually there is only so much that can be done.

2) Regarding #230, making the python parts faster makes the creation of many maps to choose from more feasible, but eventually platec will be the bottleneck and this is never going to be fun. I think @PeteMichaud is right on that issue: Regarding a world's potential to be interesting it matters far more that the shape changes with resolution than that everything else changes. So my suggestion would be to support running platec at a lower resolution than everything else and upscale the result. So you could indeed create a large number of worlds in a small resolution, select the interesting ones and rerun them with the shapes upscaled and subsequent steps running at full resolution. That way you'll get a world that has the same shape as the low-res one but is different in the details. I think this is a good tradeof.

- Somewhat related: #80, but what I mean is more than scaling after everything was completed.

3) Regarding the number of plates obviously we would want to make that an option (#215). I understand @sara-02's question in #227 to go in that direction. However, there are limitations. I don't think a "realistic" number of 50 plates, as discussed there, would work, even on the largest currently feasible map sizes. This makes me question two things:

- For what size is the current default of 10 optimized? @psi29a @ftomassetti ?
- More generally, how should the relationship between map size and numbers of plates be? Making this an option is fine, but the default should probably depend (logarithmically?) on the size of the map. For very small maps I think even 10 plates is too much.

Cheers
Alex

ftomassetti:
Maybe it is already clear but note that platec is written in C++ and it is very complex code we recovered from a thesis. We have polished it over time and addes a few tests but it is far from being easy to change.

Regarding using different resolutions, yes, that would be good but I think that running the plates simulation at an higher resolution give more interesting results. Running resolution at low resolution, scale and then perform the other operations (like erosion) would give probably slightly less nice results but the advantages seem great.

Regarding the number of plates, of course it should be an option but I think we should also give sensible defaults, maybe considering the resolution. The value was chosen by experimenting at different resolutions. I would say we have to experiment to find the optional link between size and number of plates

ftomassetti:
[...]

PR #249, opened 2017-10-01
Final speedup PR

MM1nd:
This is the last thing I had prepared. 

I discovered that `numpy.tolist` does the main thing of what `_to_protobuf_matrix` is trying to do sans transformation. We can also use numpy to vectorise the transformation which looks a bit cleaner. 

I also picked the extremely low hanging fruit of making the iteration order in `river_sources` cache friendly.

Again this is not that much faster ~10s for the 1024*512 map.

Here is how the final timing of 2m12s is mostly spent:

- ~16% `platec.step` (as discussed in #246)
- ~6% `irrigation.calculate` (somebody already took numpy to that problem, so there is little to be gained, I assume)
- ~6% `protobuf.internal.encoder.EncodeRepeatedField` (mostly type checks, AFAICT)
- ~5% `PNGWriter.set_pixel` (maybe a faster numpy->image format will fall out of my work on ancient map)
- ~4% `erosion.river_sources` (this was at about 6% before the change of iteration order)

And that's that.

### DIFF
--- a/worldengine/model/world.py
+++ b/worldengine/model/world.py
@@ -149,24 +149,21 @@ class World(object):
 
     @staticmethod
     def _to_protobuf_matrix(matrix, p_matrix, transformation=None):
-        for row in matrix:
+
+        m = matrix
+        if transformation is not None:
+            t = numpy.vectorize(transformation)
+            m = t(m)
+
+        for row in m:
             p_row = p_matrix.rows.add()
-            for cell in row:
-                '''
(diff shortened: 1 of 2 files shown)


psi29a:
Final? ;)

MM1nd:
Final as in for the world_generation path.
ancient_maps is a different animal.

psi29a:
Again, great work. It looks good to me,

PR #251, opened 2017-10-08
Ancient map anti alias

MM1nd:
This fails the test as discussed in #250 

The blessed image would have to be changed. It is two full minutes (~66%) faster on a 1024*512 map on my machine though.

### DIFF
--- a/worldengine/drawing_functions.py
+++ b/worldengine/drawing_functions.py
@@ -488,9 +488,6 @@ def draw_ancientmap(world, target, resize_factor=1,
     for c in range(num_channels):
             channels[c][borders] = border_color[c]
 
-    for c in range(num_channels):
-        target[:,:,c] = channels[c,:,:]
-
     if verbose:
         elapsed_time = time.time() - start_time
         print(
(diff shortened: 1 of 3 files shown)


worldengine/drawing_functions.py
```diff
@@ -490,37 +497,17 @@ def on_border(pos):
...
-            r = int(tot_r / n)
-            g = int(tot_g / n)
-            b = int(tot_b / n)
-            target[y, x] = (r, g, b, 255)
-
-        for i in range(steps):
-            _anti_alias_step()
-
-    anti_alias(1)
+    #don't anti-alias the alpha channel
```
ftomassetti: could we put a space after the sharp symbol?

worldengine/drawing_functions.py
```diff
@@ -490,37 +497,17 @@ def on_border(pos):
...
+    for c in range(num_channels-1):
+        channels[c] = anti_alias_channel(channels[c], 1)
+
+    
+    #hand over to the old implementation
```
ftomassetti: same here

ftomassetti:
The code looks good to me. I would first merge the changes into the blessed images (PR to worldengine-data) and then rerun the tests here. As they pass I would merge this PR.
Does it make sense?

psi29a:
I agree, if you would like to update the blessed images first, we can compare the differences there too.

psi29a:
Restarting build...

psi29a:
Coveralls is not happy... but the tests pass. :)

psi29a:
Merged, this is great work. Thank you for taking the time!

Please take the time to update the README Contributors section to add your own blurb about yourself. :)

PR #252, opened 2017-10-08
improved mountain mask

MM1nd:
In absolute timing this takes about 10s off the clock for a 1024*512 map. Relative timings are dominated by anti-aliasing, but I thought I'd do this while waiting for #251 to be resolved. 

Code is also a bit cleaner, once again.

Not much more to be said about this one.


psi29a:
Nice work! Those nested loops are gone which is a big win for simplicity. Thank you again. :)

PR #253, opened 2017-10-10
Refactor ancient map

MM1nd:
Hi again,

this more or less completely refactors `draw_ancient_map`. It is part of my effort to get rid of the slow `*_tiles_around_*` functions. 

## Changes #
Almost nothing is where it used to be, again to the point where the diff is almost useless, so let me highlight some changes:

* `find_land_borders` and `find_outer_borders` both use the scaled ocean. So it would be nice if they were in the same scope as to share that. They also both can be expressed in very few lines of code. So I moved them into `draw_ancientmap` proper.
* biome drawing code looked a bit differently for each biome but in the end only three different concepts were used. The new `_draw_biome` caters almost all our biome drawing needs. I'll try to get rid of the glacier special case next, btw.
* this results in verbose mode actually being verbose. 
* masks can be unset with absolutely no hassle using numpy array adressing syntax. No need for custom code there.
* most biomes want to know how many neighbours of them within a radius of 6 or 9 are borders. So I'm caching those values for reference in all biomes. Mountains on the other hand don't know the radius they are going to have in advance so I'm caching borders for every radius that appears, regardless of if it will ever be reused. This is wasteful but not slower than the old `*_tiles_around_*` solution. It gets more advantageous the more mountains are on a map.

## Stats #
Performance of `ancient_map -w seed_7577.world -v --draw-outer-border`: 
* Current master: 1m18s, ~14% of which in `tiles_around_factor`.
* This PR: 1m 11s (i. e. no measurable difference)

Basically this does the same thing equally fast looking better and using ~200 fewer lines of code.

Cheers,
Alex

### DIFF
--- a/tests/drawing_functions_test.py
+++ b/tests/drawing_functions_test.py
@@ -17,6 +17,13 @@ class TestDrawingFunctions(TestBase):
         draw_ancientmap(self.w, target, resize_factor=3)
         self._assert_img_equal("ancientmap_28070_factor3", target)
 
+    def test_draw_ancient_map_outer_borders(self):
+        # TODO: So far this only calls the function without testing the result.
+        # Add a blessed image with borders, maybe
+        target = PNGWriter.rgba_from_dimensions(self.w.width * 3, self.w.height * 3)
+        draw_ancientmap(self.w, target, resize_factor=3, draw_outer_land_border=True)
+
+
     def test_gradient(self):
         self._assert_are_colors_equal((10, 20, 40),
                                       gradient(0.0, 0.0, 1.0, (10, 20, 40), (0, 128, 240)))
--- a/worldengine/drawing_functions.py
+++ b/worldengine/drawing_functions.py
@@ -402,17 +402,25 @@ def draw_ancientmap(world, target, resize_factor=1,
     if draw_biome:
         biome_masks = _build_biome_group_masks(world, resize_factor)
 
-        # TODO: there was a stub for a rock desert biome group here
-        # it should be super easy to introduce that group with the new
-        # biome group concept but since it did nothing I removed the stub
+        def _draw_biome(name, _func, w, h, r, _alt_func = None):
+            if verbose:
+                start_time = time.time()
 
-    def unset_mask(pos):
-        x, y = pos
-        mountains_mask[y, x] = 0
+            for y in range(resize_factor * world.height):
+                for x in range(resize_factor * world.width):
+                    if biome_masks[name][y, x] > 0:
+                        if r == 0 or border_neighbours[r][y,x] <= 2:
+                            if _alt_func is not None and rng.random_sample() > .5:
+                                _alt_func(target, x, y, w, h)
+                            else:
+                                _func(target, x, y, w, h)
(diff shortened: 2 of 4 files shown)


tests/drawing_functions_test.py
```diff
@@ -17,6 +17,13 @@ def test_draw_ancient_map(self):
         draw_ancientmap(self.w, target, resize_factor=3)
         self._assert_img_equal("ancientmap_28070_factor3", target)
 
+    def test_draw_ancient_map_outer_borders(self):
```
ftomassetti: Good idea

worldengine/drawing_functions.py
```diff
@@ -251,7 +209,7 @@ def _draw_glacier(pixels, x, y):
     pixels[y, x] = (rg, rg, 255, 255)
 
 
-def _draw_cold_parklands(pixels, x, y):
+def _draw_cold_parklands(pixels, x, y, w, h):
```
ftomassetti: I am wrong or w and h are not used here? Are they passed for consistency with other methods?

worldengine/drawing_functions.py
```diff
@@ -411,57 +369,58 @@ def draw_ancientmap(world, target, resize_factor=1,
...
-        # biome group concept but since it did nothing I removed the stub
-
-    def unset_mask(pos):
-        x, y = pos
-        mountains_mask[y, x] = 0
-
-    def unset_boreal_forest_mask(pos):
-        x, y = pos
-        biome_masks['boreal forest'][y, x] = 0
+    scaled_ocean = world.ocean.repeat(resize_factor,0).repeat(resize_factor,1)
```
ftomassetti: super nitpick (shame on me), maybe you could put a space after the commas and between the `>` sign below

MM1nd:
Coveralls metric is quite stupid here. I'm "deleting" code so of course relative coverage goes down. 

The problem is the unused, unfinished(?) untested and uncalled `_dynamic_draw_a_mountain`, which actually would be better off in a branch.

worldengine/drawing_functions.py
MM1nd: Exactly. IIRC it was that way, then I removed them in the first round of refactoring, and now I put them back in so that duck typing works. cf. `_draw_biome`, this gets called as `_func`.

Could be solved with an ìf` but branching is slow.

ftomassetti: makes sense

worldengine/drawing_functions.py
MM1nd: OK, I'm giving up and installing pylint :)

psi29a: PyCharm IDE is also free... unless you're an vim or emacs fan, then yeah, pylint and pep8 are your friends. :)

MM1nd: I'm using Visual Studio and don't want to change my workflow, if I can help it. So kindof the opposite of vim :)
It has built-in formatting I think, but that has to be triggered manually, which I maybe just should remember to do. It also has builtin pylint support which I wanted to give a try ... mixed results so far.

MM1nd:
I fixed the whitespaces as requested, I also did a lot of linting and will make a seperate PR for that after this one is done.

PR #255, opened 2017-10-15
Linting

MM1nd:
Some improvements to the looks of the code.


psi29a:
This looks OK to me, I have a few... 'meh... why?' moments, but nothing that I'm like 'nonononono'.
@ftomassetti any comments?

tests/draw_test.py
```diff
@@ -44,42 +44,45 @@ def _assert_img_equal(self, blessed_image_name, drawn_image):
                         "and/or amount of channels. Blessed %s, drawn %s"
                         % (str(blessed_img.array.shape), str(drawn_image.array.shape)))
 
-        # compare images; cmp_array will be an array of booleans in case of equal shapes (and a pure boolean otherwise)
+        # compare images;
```
ftomassetti: should be ":" instead of ";"?

worldengine/biome.py
```diff
@@ -270,8 +270,9 @@ class TropicalRainForest(Biome, Jungle):
 
 def biome_name_to_index(biome_name):
     names = sorted(_BiomeMetaclass.biomes.keys())
-    for i in range(len(names)):
-        if names[i] == biome_name:
+
```
ftomassetti: maybe we do not need this new extra line

worldengine/draw.py
```diff
@@ -426,7 +431,7 @@ def draw_satellite(world, target):
                             all_b.append(b)
 
                 # Making sure there is at least one valid tile to be smoothed before we attempt to average the values
-                if len(all_r) > 0:
+                if all_r:
```
ftomassetti: are these equivalent?

worldengine/generation.py
```diff
@@ -142,56 +150,55 @@ def harmonize_ocean(ocean, elevation, ocean_level):
...
-    # We want to multiply the raw sea_depth by one of these factors 
+    # We want to multiply the raw sea_depth by one of these factors
     # depending on the distance from the next land
     # possible TODO: make this a parameter
     factors = [0.0, 0.3, 0.5, 0.7, 0.9]
 
     next_land = next_land_dynamic(world.layers['ocean'].data)
 
-    sea_depth = sea_level - world.layers['elevation'].data
+    result = sea_level - world.layers['elevation'].data
```
ftomassetti: I do not like the name result much here

worldengine/model/world.py
```diff
@@ -312,17 +313,19 @@ def _from_protobuf_world(cls, p_world):
...
-        if len(p_world.biome.rows) > 0:
-            w.biome = numpy.array(World._from_protobuf_matrix(p_world.biome, biome_index_to_name), dtype=object)
+        if p_world.biome.rows:
+            w.biome = numpy.array(
+                World._from_protobuf_matrix(p_world.biome, biome_index_to_name), 
+                dtype=object)
 
         # Humidity
-        if len(p_world.humidity.rows) > 0:
+        if p_world.humidity.rows:
```
ftomassetti: is this equivalent?

MM1nd:
"This looks OK to me, I have a few... 'meh... why?' moments, but nothing that I'm like 'nonononono'."

The answer to "why?" is almost certainly "because pylint said so" and in most cases likely "because the line was too long". I'm agnostic about all of this as witnessed by the three "whitespace" commits at the end of every of my PRs :)

tests/draw_test.py
MM1nd: Sure.

worldengine/biome.py
MM1nd: Sure.

worldengine/draw.py
MM1nd: cf. https://stackoverflow.com/questions/43121340

worldengine/generation.py
MM1nd: sea_depth is the name of the function and pylint rightfully complains that this redefines a name from the outer scope. Personally I use `result` a lot as a variable name, but I'm open to anything that is not `sea_depth`. Alternatively the function needs a new name.

worldengine/model/world.py
MM1nd: cf. above

worldengine/generation.py
psi29a: depth_of_sea  (in the very french way of saying sea_depth)

psi29a:
I usually set my columns to 120 (from 80)  and adjust linting accordingly...

ftomassetti:
Me too, I use 120 columns and that settings is suggested [here](https://github.com/Mindwerks/worldengine/wiki/Coding-guidelines) (arguably very difficult to find)

PR #260, opened 2019-11-16
Add GDAL dependencies to Dockerfile

nhynes:
This PR adds the libraries needed for GDAL support to the Dockerfile, which itself has been moved to a directory so to keep the build context from picking up development/output files.


ftomassetti:
The build seems to fail on Travis. Could you take a look?

psi29a:
Numpy gives:
>  RuntimeError: Python version >= 3.5 required.

nhynes:
It looks like some transitive dependencies in the main package have become stale. Changing the dockerfile wouldn't have caused that. If anything, dockerized tests would have prevented the problem entirely :)

On one hand, I'd be glad to churn through the CI, but it might actually be easier to write a Rust version for high-res worlds. This code and the research behind it is solid, though!

ftomassetti:
If you are willing to fix this we would be very grateful :)
I understand how you think that rewrite this in Rust would be easier but in my opinion this could prove to take a little longer than you expect...
In the long run it would be good to have this project in a language easier to package however there are other aspects to consider.
1) More people understand Python
2) This project has some popularity
3) There is a vast amount of knowledge in this codebase. Rewriting it and get everything exactly right will be a very significant effort in my opinion

PR #261, opened 2022-02-17
Increase numpy version

ctittel:
The older numpy versions that were in setup.py failed to install on both systems I tested it on (Ubuntu 20.04 and EndeavourOS).
I bumped the numpy version much higher and now it works.

(This shouldn't affect the tests but some tests failed, maybe because of changes with Python's abstract class module)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ config = {
     'entry_points': {
         'console_scripts': ['worldengine=worldengine.cli.main:main'],
     },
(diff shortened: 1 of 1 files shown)


ftomassetti:
This project is currently unmaintained. We are looking for volunteers to maintaining it

akaszynski:
@ftomassetti, please ping me regarding maintenance. I have a small amount of time to spare to keep this up-to-date.

ftomassetti:
@akaszynski if @psi29a agrees we could then make you the maintainer of the project

psi29a:
Already done in #261

